Issue #21, opened 2026-02-01
Add feature to install/update new plugins without app update


robinebers:
Thinking settings tab for now, load a ZIP or something

Maybe load from URL

Maybe later from a dedicated GitHub repository, incl. discovery or auto detect

PR #84, opened 2026-02-06
fix: tighten security controls (CSP, plugin permissions, pinned deps, audit lockfile)

aravhawk:
# Motivation
- Harden plugin runtime and frontend surface; reduce XSS and plugin-driven exfiltration risk; restore dependency auditability.

# Description
- Enable restrictive CSP in Tauri config to replace `csp: null`.
- Add manifest-driven plugin permissions and persist on `LoadedPlugin`.
- Enforce allowlists for fs/http/keychain/sqlite host APIs with path normalization and host gating.
- Thread permissions into runtime host API injection.
- Update plugin manifests with explicit permissions (claude/codex/copilot/cursor/mock).
- Pin tauri-nspanel git dependency and add package-lock + fixes spec for auditability.

# Testing
- `npm install --package-lock-only --ignore-scripts`
- `npm audit`

---
## Summary by cubic
Enforces a strict CSP and a manifest-based permission model for plugins to limit filesystem, network, keychain, and SQLite access. Pins tauri-nspanel to a specific commit and adds package-lock.json to enable npm audit.

- New Features
  - Strict CSP in Tauri (no inline scripts; dev sockets to localhost:1420 allowed).
  - Plugins must declare permissions; host APIs enforce allowlists for fs paths, HTTP hosts, keychain services, and SQLite db paths.
  - Path tokens supported: appDataDir, pluginDataDir; HTTP supports wildcard hosts (e.g., *.example.com) and blocks non-HTTPS unless explicitly allowed.

- Migration
  - Update each plugin’s plugin.json with permissions (fs.read/write, http.allowedHosts, keychain.services, sqlite.allowedPaths).
  - For non-HTTPS calls, set http.allowHttp: true (use sparingly).
  - If the UI relies on inline scripts/styles, adjust to comply with the new CSP or add targeted exceptions.

<sup>Written for commit 807b61223ebe07179b52db135feecbf5f8a212cc. Summary will update on new commits.</sup>

### DIFF
--- a/docs/security-audit-2026-02-06.md
+++ b/docs/security-audit-2026-02-06.md
@@ -1,0 +1,66 @@
+# Security audit (2026-02-06)
+
+## Executive summary
+- Overall risk: **High** due to disabled CSP and un-sandboxed plugin execution with privileged host APIs.
+- No evidence of obvious credential leakage in logs; redaction exists in host API logging.
+- Dependency audit tooling incomplete because npm audit requires a package-lock; process gap remains.
+
+## Scope
+- Static review of Tauri configuration, plugin execution model, and privileged host APIs.
+- Dependency risk review from `package.json` and Rust crate configuration.
+- Targeted pattern scan for dangerous APIs in application source.
+
+## Methods (local)
+- Manual config and code review (`src-tauri/*`, `src/*`, `plugins/*`).
+- Targeted pattern scans for CSP and unsafe API usage.
+- Attempted `npm audit` (blocked by missing package-lock).
+
+## Findings
+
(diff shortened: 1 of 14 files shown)


robinebers:
Thank you, Arav! Very valuable contribution that I need to thoroughly test as part of #8.

Right now, this is not a huge security issue because the plugins are bundled and rewritten on every launch (was just easier this way, while using the real plugin-loading flow). This will proof valuable once I start working on #21.

Keeping this open in the meantime.

robinebers:
This will be the first priority as in one of the next versions (should be the next big core update), but closing for now. Thank you so much for it, I will reference it heavily and credit accordingly.

Issue #85, opened 2026-02-06
Refresh data when click "Next update in " label


Guria:
Currently to manually trigger refresh I need to go settings and toggle checkbox on provider.
Would be nice to call it manually on click.

Alternatively / Additionally may add the setting to auto update data when users clicks on menu bar icon as it most likely indicates user intent to know fresh data

robinebers:
you can manually refresh by hovering over the provider title, there’s a little hidden refresh button that pops up

Guria:
Oh, wow. Totally missed it. I think we need to increase the hover area that triggers this button to appear.

Right now, I have to hover exactly over the spot where the button is hidden, and it’s almost impossible to discover without prior knowledge. It would help if hovering over the provider name title text area revealed the button too.

That said, I still think the “Next update in” label is a better place for this action, since it is more directly related to updating. Maybe we could use the same appear-on-hover icon style there, but making the entire label clickable would probably be the most intuitive way to surface this action.

robinebers:
It used to be there, I can bring it back. It will trigger a cooldown though, so people don't spam it.

robinebers:
@cursoragent we used to have this functionality. Look through the Comets from a few days ago and see if you can restore the core feature. Essentially when clicking on the next update label, we want to trigger a global refresh that triggers a cool down for every refreshed plugin.

PR #112, opened 2026-02-07
Add 'Refresh All' cooldown and UI

muskecan:
Prevent repeated "Refresh All" spamming by adding a cooldown. App.tsx: introduce REFRESH_COOLDOWN_MS usage, track lastRefreshAllAt state, implement handleRefreshAll to respect cooldown, mark plugins as manually refreshing, start batch refresh and set errors on failure. Settings.tsx: expose onRefreshAll and lastRefreshAllAt props, add tooltip/hourglass/refresh icons and a conditional Refresh button that shows remaining cooldown (uses useNowTicker), and import necessary UI components. This UX change shows remaining time and disables triggering until the cooldown expires.

## Description

Adds a manual refresh button to the Auto Refresh settings section. The button sits next to the "Auto Refresh" heading as a small circular arrow icon (RefreshCw) that appears on hover — matching the exact same interaction pattern used on individual provider cards.

- Clicking the button immediately refreshes all enabled plugins, bypassing the current auto-refresh timer
- The auto-refresh countdown resets after the manual trigger (so the next auto-refresh starts fresh from the full interval)

Cooldown:

- Uses the same 5-minute cooldown (REFRESH_COOLDOWN_MS) already in place for per-provider refresh buttons
- During cooldown the icon swaps to an Hourglass, dimmed and non-clickable
- Hovering the hourglass shows a tooltip with a live countdown ("Available in 4m 32s")
- The handler itself also guards against cooldown bypass (early-return if within window)

Before 

image

After

image

## Type of Change

- [ ] Bug fix
- [X] New feature
- [ ] New provider plugin
- [ ] Documentation
- [ ] Performance improvement
- [ ] Other (describe below)

## Testing

- [X] I ran `bun run build` and it succeeded
- [X] I ran `bun run test` and all tests pass
- [X] I tested the change locally with `bun tauri dev`

## Checklist

- [X] I read [CONTRIBUTING.md](../CONTRIBUTING.md)
- [X] My PR targets the `main` branch
- [X] I did not introduce new dependencies without justification

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -31,6 +31,7 @@ import {
   loadTrayIconStyle,
   loadThemeMode,
   normalizePluginSettings,
+  REFRESH_COOLDOWN_MS,
   saveAutoUpdateInterval,
   saveDisplayMode,
   savePluginSettings,
@@ -81,6 +82,7 @@ function App() {
 
   const { updateStatus, triggerInstall } = useAppUpdate()
   const [showAbout, setShowAbout] = useState(false)
+  const [lastRefreshAllAt, setLastRefreshAllAt] = useState<number | null>(null)
 
   const trayRef = useRef<TrayIcon | null>(null)
   const trayGaugeIconPathRef = useRef<string | null>(null)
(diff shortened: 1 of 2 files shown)


robinebers:
This is great, thanks. Instead of putting the refresh button in the settings, #85 suggested to make the "next update" label (which is visible everywhere) clickable.

**There's also:**
- Missing test coverage
- Duplicated logic (essentially copy-pasted)
- Missing loading UX (would be solved by moving it to the update label)

PR #134, opened 2026-02-09
feat: add 1-minute refresh interval and replace 1-hour option

### LINKED ISSUE #133: Add 1-minute refresh interval
### Problem

The current minimum refresh interval in OpenUsage is 5 minutes.

Based on my personal experience, this can feel a bit long when trying to understand usage changes more immediately, especially during periods of frequent requests.

### Proposed Solution

One possible improvement could be adding an optional 1-minute refresh interval for more immediate usage feedback during active development.

From a personal perspective, a 1-hour refresh interval feels relatively long for a usage-tracking tool like OpenUsage. From a UI simplicity standpoint, it might be worth considering replacing the 1-hour option with a 1-minute option.

That said, this is just a personal suggestion, and I’ll leave the final decision to the maintainer.

### Alternatives Considered

_No response_

### Additional Context

_No response_

hanseo0507:
## Summary
- Add a 1-minute auto-update interval option for more immediate usage feedback during active development
- Remove the 1-hour interval option in favor of the new 1-minute option (intervals: 1, 5, 15, 30 min)
- Reduce manual refresh cooldown from 5 minutes to 1 minute to align with the new minimum interval

Closes #133

## Test plan
- [ ] Verify the settings UI shows "1 min", "5 min", "15 min", "30 min" options
- [ ] Verify selecting "1 min" triggers auto-refresh every minute
- [ ] Verify manual refresh cooldown is ~1 minute

### DIFF
--- a/src/App.test.tsx
+++ b/src/App.test.tsx
@@ -906,9 +906,9 @@ describe("App", () => {
     await userEvent.click(settingsButtons[0])
 
     // Change interval - this triggers the if branch (enabledIds.length > 0)
-    await userEvent.click(await screen.findByRole("radio", { name: "1 hour" }))
+    await userEvent.click(await screen.findByRole("radio", { name: "1 min" }))
 
-    expect(state.saveAutoUpdateIntervalMock).toHaveBeenCalledWith(60)
+    expect(state.saveAutoUpdateIntervalMock).toHaveBeenCalledWith(1)
   })
 
   it("fires auto-update interval and schedules next", async () => {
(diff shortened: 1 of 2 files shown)


robinebers:
Hi @hanseo0507 

Thank you

I decided not to support this, because it creates a lot of unintentional requests that these providers do not expect, and have to deal with because of this this and similar tools. This change would pretty much 5x the number of requests, which I don't want to cause.

Hope that makes sense!

PR #154, opened 2026-02-10
fix(panel): improve tray icon positioning logic for macOS

### LINKED ISSUE #150: Wrong position on secondary screen
### Description

When using an extra monitor and opening Openusage from the toolbar of this monitor, the overlay is displayed in the top-right corner of the primary screen.

### Steps to Reproduce

1. On an extra monitor click on the OpenUsage toolbar icon

### Expected Behavior

The app overlay should appear on the screen with the icon clicked

### Actual Behavior

The overlay appears on the primary macbook screen

### OpenUsage Version

0.6.0

### Operating System

macOS (Apple Silicon)

### Screenshots

Screenshot from top-right corner of my primary screen (macbook)

Image

### Additional Context

_No response_

MuhammadAli511:
## Description
Fix panel overlay appearing on primary monitor instead of the monitor where the tray icon was clicked.

## Related Issue
Fixes #150 

## Type of Change
- [x] Bug fix
- [ ] New feature
- [ ] New provider plugin
- [ ] Documentation
- [ ] Performance improvement
- [ ] Other (describe below)

## Testing

## Screenshots
Before:
Screenshot 2026-02-11 at 2 47 52 AM

After:
Screenshot 2026-02-11 at 2 34 59 AM

## Checklist

---
## Summary by cubic
Fixes the macOS panel opening on the wrong display by anchoring it to the monitor under the cursor, with consistent behavior on mixed‑DPI setups. Also prevents edge-case misplacement on monitor seams.

- **Bug Fixes**
  - Uses NSEvent::mouseLocation (via objc2-app-kit) in unified point space to pick the correct monitor; flips to top‑left origin, falls back to primary if needed, and uses half‑open bounds to avoid seam double matches.
  - Converts the tray icon’s physical position and size to that monitor’s points, centers the panel under the icon with a 6pt nudge, reads the panel width from the window (or tauri.conf.json), and sets the position using LogicalPosition to avoid mixed‑DPI drift.

- **Refactors**
  - Removed leftover debug logs.

<sup>Written for commit df7bd5189b03c295384b3f3327aa200eecc5beb1. Summary will update on new commits.</sup>

### DIFF
--- a/src-tauri/src/panel.rs
+++ b/src-tauri/src/panel.rs
@@ -160,8 +160,8 @@ pub fn position_panel_at_tray_icon(
 
         let logical_x = pos.x as f64 / scale;
         let logical_y = pos.y as f64 / scale;
-        let x_in = mouse_x >= logical_x && mouse_x <= logical_x + logical_w;
-        let y_in = mouse_y >= logical_y && mouse_y <= logical_y + logical_h;
+        let x_in = mouse_x >= logical_x && mouse_x < logical_x + logical_w;
+        let y_in = mouse_y >= logical_y && mouse_y < logical_y + logical_h;
 
         if x_in && y_in {
             found_monitor = Some(m.clone());


robinebers:
Thank you so much! I can’t easily test it for the next two days. But if others jump in and review this, I’d be happy to merge it.

robinebers (commented):
Hey! 👋 This is Rob's AI reviewer. Thanks for the contribution — really appreciate you tackling this multi-monitor bug!

The approach is solid: the coordinate model (logical_pos × scale = physical origin) correctly matches the upstream Tauri bug (tauri-apps/tauri#7890), the fallback to primary monitor is much safer than the old `expect()` panic, switching to `LogicalPosition` is the right call, and the inline comments explaining *why* are excellent. Two things to address before merge:

1. **Icon-width heuristic breaks for dynamic tray icons** — The `icon_phys_w > 100.0` threshold was likely tested against the static fallback PNG (~132px on Retina). But OpenUsage dynamically replaces the tray icon with variable-width rendered content (bars, circles, text like "75%"). On a 2x display, most of these produce icons well under 100px — for example, bars-only is ~36px, bars + "75%" is ~99px, circle is ~36px. That means the scale detection silently fails for the majority of real-world states, and the overlapping-region disambiguation (the whole point of the fix) won't work. A more robust approach: instead of inferring scale from icon size, compare `icon_phys_w` against the monitor's own scale factor when disambiguating. For example, check if `(icon_phys_w / expected_logical_w)` is closer to each candidate monitor's scale, or simply use the monitor whose scale factor best explains the icon's reported physical dimensions.

2. **Hardcoded panel width** — `let panel_width = 400.0_f64` duplicates a value from `tauri.conf.json`. The old code read this from `window.outer_size()`, which stayed in sync automatically. If the configured width ever changes, this silently breaks positioning. Consider using `window.outer_size()` again, converting to logical points.

Everything else — the monitor detection logic, the coordinate math, graceful fallback, and the quality of the comments — looks great. Nice work!

MuhammadAli511:
Looking into it.

MuhammadAli511:
@robinebers Fixed the PR comments.

PR #200, opened 2026-02-20
feat(tray): use selected provider for menubar icon and percentage

yaacov9:
## Summary
I’m a CodexBar user who switched to OpenUsage, and this behavior was one feature I wanted badly in day-to-day use.

Before this change, the menubar icon and percentage were always tied to the first enabled provider in Settings. That felt confusing when I was actively looking at another provider in the left sidebar.

This PR makes the tray follow the provider currently selected in the sidebar. If no specific provider is selected (Home/Settings), behavior stays exactly as today and falls back to the first enabled provider.

Built and tested by me with help from GPT-5.3-Codex.

## What changed
- Added optional `preferredPluginId` support in tray primary progress calculation.
- Wired selected sidebar provider (`activeView`) into tray icon/percentage source selection.
- Preserved fallback behavior when no provider is selected.
- Added regression tests:
  - tray logic prioritization + fallback tests
  - app-level test that selecting a provider updates provider tray icon source

## Workflow improvement
- Menubar now reflects the provider you are currently inspecting, which removes the “why am I seeing Cursor while I’m on Codex?” mismatch.
- The existing default remains unchanged when nothing is selected, so this improves clarity without changing established behavior for users who stay on Home/Settings.

## Validation
- `npm run build`
- `npm test -- src/lib/tray-primary-progress.test.ts src/App.test.tsx`
- `npm run test:coverage` (fails on existing repo threshold in `plugins/gemini/plugin.js` branch coverage; unrelated to this change)

---
## Summary by cubic
The tray menubar icon and percentage now follow the provider selected in the sidebar, and updates are requeued if a render is in flight to avoid stale display. If no provider is selected or it’s disabled/unknown, it falls back to the first enabled provider.

- **Bug Fixes**
  - Added preferredPluginId to prioritize the selected, enabled provider.
  - Wired activeView into tray rendering and trigger updates when selection changes.
  - Requeue tray updates during in-flight renders so the latest selection is applied.
  - Preserved fallback behavior when no provider is selected.
  - Added tests for provider icon selection, prioritization/fallback, and requeue behavior.

<sup>Written for commit 8fb5d2a99cd4be02c71f79fab96468ad0c077b55. Summary will update on new commits.</sup>

### DIFF
--- a/src/App.test.tsx
+++ b/src/App.test.tsx
@@ -502,6 +502,57 @@ describe("App", () => {
     })
   })
 
+  it("requeues selected provider tray update while a tray render is in flight", async () => {
+    state.loadTrayIconStyleMock.mockResolvedValueOnce("provider")
+    state.invokeMock.mockImplementation(async (cmd: string) => {
+      if (cmd === "list_plugins") {
+        return [
+          {
+            id: "a",
+            name: "Alpha",
+            iconUrl: "icon-a",
+            primaryCandidates: ["Session"],
+            lines: [{ type: "progress", label: "Session", scope: "overview" }],
+          },
+          {
+            id: "b",
+            name: "Beta",
+            iconUrl: "icon-b",
+            primaryCandidates: ["Session"],
+            lines: [{ type: "progress", label: "Session", scope: "overview" }],
+          },
+        ]
+      }
+      return null
+    })
+    state.loadPluginSettingsMock.mockResolvedValueOnce({ order: ["a", "b"], disabled: [] })
+
+    let resolveFirstRender: ((img: unknown) => void) | null = null
+    state.renderTrayBarsIconMock
+      .mockImplementationOnce(
+        () =>
+          new Promise((resolve) => {
+            resolveFirstRender = resolve
+          })
+      )
+      .mockResolvedValue({})
+
+    render(<App />)
+    await waitFor(() => expect(state.renderTrayBarsIconMock).toHaveBeenCalledTimes(1))
+
+    await userEvent.click(await screen.findByRole("button", { name: "Beta" }))
+    expect(state.renderTrayBarsIconMock).toHaveBeenCalledTimes(1)
+
+    resolveFirstRender?.({})
+
+    await waitFor(() => {
+      expect(state.renderTrayBarsIconMock.mock.calls.length).toBeGreaterThan(1)
+      const latestCall = state.renderTrayBarsIconMock.mock.calls.at(-1)?.[0]
+      expect(latestCall.bars[0]?.id).toBe("b")
+      expect(latestCall.providerIconUrl).toBe("icon-b")
+    })
+  })
+
   it("uses text-only tray style", async () => {
     state.loadTrayIconStyleMock.mockResolvedValueOnce("textOnly")
     state.invokeMock.mockImplementation(async (cmd: string) => {
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -109,6 +109,7 @@ function App() {
   const trayGaugeIconPathRef = useRef<string | null>(null)
   const trayUpdateTimerRef = useRef<number | null>(null)
   const trayUpdatePendingRef = useRef(false)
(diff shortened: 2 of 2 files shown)


src/App.test.tsx
```diff
@@ -459,6 +459,49 @@ describe("App", () => {
     expect(latestCall.percentText).toBe("50%")
   })
 
+  it("uses selected provider for provider tray style when selecting from sidebar", async () => {
```
davidarny, line 462: Nice coverage on provider switching. One gap remains: this test never exercises the in-flight tray update path. `renderTrayBarsIcon` resolves immediately here, so `trayUpdatePendingRef` never blocks a queued update.

Please add a regression where the first `renderTrayBarsIcon` call stays pending, then click `Beta`, resolve the first promise, and assert we get a follow-up render with `bars[0].id === "b"` and `providerIconUrl === "icon-b"`.

A simple pattern:

```ts
let resolveFirst: ((img: unknown) => void) | null = null
state.renderTrayBarsIconMock
  .mockImplementationOnce(() => new Promise((res) => {
    resolveFirst = res
  }))
  .mockResolvedValue({})

render(<App />)
await waitFor(() => expect(state.renderTrayBarsIconMock).toHaveBeenCalledTimes(1))

await userEvent.click(await screen.findByRole("button", { name: "Beta" }))
resolveFirst?.({})

await waitFor(() => {
  const latest = state.renderTrayBarsIconMock.mock.calls.at(-1)?.[0]
  expect(latest.bars[0]?.id).toBe("b")
  expect(latest.providerIconUrl).toBe("icon-b")
})
```

src/App.tsx
```diff
@@ -156,6 +157,7 @@ function App() {
         pluginStates: pluginStatesRef.current,
         maxBars,
         displayMode: displayModeRef.current,
+        preferredPluginId: selectedTrayProviderIdRef.current,
```
davidarny, line 171: Related race fix idea: if a tray update is already in flight, requeue it instead of dropping it.

At the top of the timer callback, `if (trayUpdatePendingRef.current) return` can lose a provider-switch update. Consider setting `trayUpdateQueuedRef.current = true` there. Then in each `finally` block, if queued is true, clear it and call `scheduleTrayIconUpdate("settings", 0)`.

This keeps debounce behavior and still guarantees the latest selected provider reaches the tray.

yaacov9:
Implemented both requested changes in commit 8fb5d2a:

1. Added queueing for tray updates while a render is in-flight (`trayUpdateQueuedRef`) so provider-switch updates are not dropped.
2. Added a regression test that keeps the first `renderTrayBarsIcon` promise pending, switches to Beta, resolves, and verifies follow-up render uses Beta (`bars[0].id` + `providerIconUrl`).

Validation run locally:
- `npm run test -- src/App.test.tsx`
- `npm run test -- src/lib/tray-primary-progress.test.ts`
- `npm run build`

davidarny:
@robinebers I personally don't look at the menu bar icon at all (perhaps because of my poor eyesight and the small size of the icons in the menu bar). I always click on the icon and open the panel to gather information. 

Perhaps it would be worth conducting a survey to find out how much demand there is for this functionality? If, for example, most people, like me, don't look at the statistics in the icon, then maybe it would be worth removing this functionality altogether?

robinebers:
> @robinebers I personally don't look at the menu bar icon at all (perhaps because of my poor eyesight and the small size of the icons in the menu bar). I always click on the icon and open the panel to gather information.
> 
> Perhaps it would be worth conducting a survey to find out how much demand there is for this functionality? If, for example, most people, like me, don't look at the statistics in the icon, then maybe it would be worth removing this functionality altogether?

I am always a bigger fan of just making the change and seeing how people react, because if we make the change and a lot of people complain we made a mistake, we roll back and think about a different solution. If no one complains, then we made the right call. 

I would probably remove the icon styles and run with a simpler version.

robinebers:
Reworked and superseded in #215, so we’re closing this one to keep discussion in the updated PR. Credit to @yaacov9 for the original work here.

PR #209, opened 2026-02-22
refactor(app): split monolithic App into focused hooks and atomic stores

davidarny:
## Problem
`src/App.tsx` had become a large mixed-responsibility file (panel lifecycle, tray rendering, probe orchestration, settings lifecycle, and view routing/rendering), which made behavior-safe changes hard to reason about and risky to review.

## Solution
- Split logic into focused domain hooks under `src/hooks/app` (probe, tray, panel, and concrete settings hooks).
- Reduced `src/App.tsx` to a composition root that wires state + hooks.
- Extracted UI composition into `src/components/app/app-shell.tsx` and `src/components/app/app-content.tsx`.
- Introduced small atomic Zustand stores in `src/stores` (`app-ui-store`, `app-plugin-store`, `app-preferences-store`, `app-derived-store`).
- Removed god-mode orchestration patterns and wired concrete settings hooks directly.

## Validation
- `bun run test -- src/App.test.tsx`
- `bun run test`
- `bun run build`

---
## Summary by cubic
Split the monolithic App into focused hooks and small Zustand stores; App is now a thin composition root with AppShell/AppContent. Removed derived-state mirroring and tightened panel, tray, and version lifecycles with docs and regression tests.

- **Refactors**
  - Added domain hooks: panel; probe (state/refresh/auto-update); tray icon; plugin views; settings (bootstrap, theme, display/system/tray/plugin actions, plugin list); app version.
  - Introduced stores: app-ui, app-preferences, app-plugin; App passes derived plugin view data directly into AppShell/AppContent.
  - Documented app state flow (app-state-architecture.md) and expanded focused tests, including store isolation; added Zustand dependency.

- **Bug Fixes**
  - Eliminated stale frames by rendering derived data in the same pass; guarded provider fallback redirects while plugin settings load with a startup test.
  - Scoped panel effects to desktop runtime; queued tray icon updates and cleaned up timers/teardown to prevent stray updates, with regression tests for teardown.
  - Hardened app version loading on failures; removed StrictMode-hostile store resets on App unmount.

<sup>Written for commit c445fa8366a64b22deb51e38b2bec420260fcb28. Summary will update on new commits.</sup>

### DIFF
--- a/README.md
+++ b/README.md
@@ -27,7 +27,7 @@ OpenUsage lives in your menu bar and shows you how much of your AI coding subscr
 - [**Claude**](docs/providers/claude.md) / session, weekly, extra usage, local token usage (ccusage)
 - [**Codex**](docs/providers/codex.md) / session, weekly, reviews, credits
 - [**Copilot**](docs/providers/copilot.md) / premium, chat, completions
-- [**Cursor**](docs/providers/cursor.md) / credits, plan usage, on-demand
+- [**Cursor**](docs/providers/cursor.md) / credits, plan usage, on-demand, CLI auth
 - [**Factory / Droid**](docs/providers/factory.md) / standard, premium tokens
 - [**Gemini**](docs/providers/gemini.md) / pro, flash, workspace/free/paid tier
 - [**Kimi Code**](docs/providers/kimi.md) / session, weekly
--- a/docs/providers/cursor.md
+++ b/docs/providers/cursor.md
@@ -105,9 +105,16 @@ Returns limit policy status plus any active credit grants. Response undocumented
 
 ## Authentication
 
-### Token Location
+### Token Sources
 
-SQLite database at `~/Library/Application Support/Cursor/User/globalStorage/state.vscdb`
+OpenUsage reads Cursor auth in this order:
+
+1. **Cursor Desktop SQLite** (preferred)
+2. **Cursor CLI keychain** (fallback)
+
+#### 1) Cursor Desktop SQLite (preferred)
+
+Path: `~/Library/Application Support/Cursor/User/globalStorage/state.vscdb`
 
 ```bash
 sqlite3 ~/Library/Application\ Support/Cursor/User/globalStorage/state.vscdb \
@@ -122,9 +129,22 @@ sqlite3 ~/Library/Application\ Support/Cursor/User/globalStorage/state.vscdb \
 | `cursorAuth/stripeMembershipType` | Plan tier (e.g. `pro`, `ultra`) |
 | `cursorAuth/stripeSubscriptionStatus` | Subscription status |
 
+#### 2) Cursor CLI keychain (fallback)
+
+OpenUsage reads Cursor CLI tokens from keychain:
+
+- `cursor-access-token`
+- `cursor-refresh-token`
+
+To initialize CLI auth:
+
+```bash
+agent login
+```
+
 ### Token Refresh
 
-Access tokens are short-lived JWTs. The app refreshes before each request if expired.
+Access tokens are short-lived JWTs. The app refreshes before each request if expired, then persists the new access token back to the same source it was loaded from (SQLite or keychain).
 
 ```
 POST https://api2.cursor.sh/oauth/token
--- a/plugins/cursor/plugin.js
+++ b/plugins/cursor/plugin.js
@@ -1,6 +1,8 @@
 (function () {
   const STATE_DB =
     "~/Library/Application Support/Cursor/User/globalStorage/state.vscdb"
+  const KEYCHAIN_ACCESS_TOKEN_SERVICE = "cursor-access-token"
+  const KEYCHAIN_REFRESH_TOKEN_SERVICE = "cursor-refresh-token"
   const BASE_URL = "https://api2.cursor.sh"
   const USAGE_URL = BASE_URL + "/aiserver.v1.DashboardService/GetCurrentPeriodUsage"
   const PLAN_URL = BASE_URL + "/aiserver.v1.DashboardService/GetPlanInfo"
@@ -9,6 +11,7 @@
   const REST_USAGE_URL = "https://cursor.com/api/usage"
   const CLIENT_ID = "KbZUR41cY7W6zRSdpSUJ7I7mLYBKOCmB"
   const REFRESH_BUFFER_MS = 5 * 60 * 1000 // refresh 5 minutes before expiration
+  const LOGIN_HINT = "Sign in via Cursor app or run `agent login`."
 
   function readStateValue(ctx, key) {
     try {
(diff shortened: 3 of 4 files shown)


davidarny:
Hey @robinebers @validatedev !

Please review this one, no need to be merged, just want to share my thoughts. So basically we have a god-mode App component, it's very messy and bloated.

In this PR I (and codex) refactored this stuff a little bit, moved some login into seperate hooks and intruduced [zustand](https://zustand-demo.pmnd.rs/) stores (instead useState).

Please, share you opinion. Do we even need this type of work or we just leave everything on the AI side?

robinebers:
this is funny because while I usually don't read code, I did today a bit and yeah this code has gotten a bit messy

totally works, but could use a cleanup

will review!

davidarny:
Thanks @robinebers, will take a look!

davidarny:
I addressed all requested follow-ups in c2ed824cfce7987aba0e5ed2e53aa5310961ccf8:

- Made derived-state plumbing explicit in `src/stores/app-derived-store.ts` (`setPluginViews({ displayPlugins, navPlugins })`).
- Added architecture note: `docs/app-state-architecture.md`.
- Added focused tests:
  - `src/hooks/app/use-app-plugin-views.test.ts`
  - `src/hooks/app/use-settings-plugin-list.test.ts`
  - `src/stores/app-derived-store.test.ts`

Validation run:
- `bun run test -- src/hooks/app/use-app-plugin-views.test.ts src/hooks/app/use-settings-plugin-list.test.ts src/stores/app-derived-store.test.ts`
- `bun run test`
- `bun run build`

validatedev:
Review from Amp:

src/App.tsx
L227: Synchronizing derived state from hooks into a Zustand store via `useEffect` causes a predictable double-render on every update: once for the hook state change and once for the store update.

Consider passing `navPlugins`, `displayPlugins`, and `autoUpdateNextAt` as direct props to `AppShell`. Since `App` is the composition root, prop drilling here is only one level deep. This approach is better than the performance and complexity costs of effect-based state synchronization.

src/hooks/app/use-app-plugin-views.ts
L57: This effect forces a redirect to "home" if `navPlugins` does not contain the `activeView`. However, `navPlugins` depends on `pluginSettings`, which loads asynchronously.

If `activeView` is persisted but `pluginSettings` is still loading (is null), `navPlugins` will be empty. Consequently, the app will immediately redirect the user to "home" before the settings have a chance to load.



Review from Claude Code (Claude Opus 4.6 High + GPT 5.3 Codex Xhigh):

PR #209 Review: Split monolithic App into focused hooks and atomic stores

  Overview

  This PR refactors the 1020-line App.tsx monolith into:
  - 4 Zustand stores (app-ui, app-plugin, app-preferences, app-derived)
  - ~14 domain hooks under src/hooks/app/
  - 2 composition components (AppShell, AppContent)

  App.tsx becomes a slim composition root that computes derived values in hooks, then mirrors them into the derived store via useEffect for child components to consume.

  Verdict: Correct with two actionable issues

  All 715 tests pass. The refactoring faithfully preserves existing behavior with two real issues worth addressing before merge.

  ---
  Issues Found

  1. [SHOULD-FIX] Derived store mirror effects can flash stale UI

  Files: src/App.tsx:206-216, src/components/app/app-content.tsx:86-89

  The derive-then-mirror pattern uses useEffect to push computed values into app-derived-store:

  useEffect(() => { setPluginViews({ displayPlugins, navPlugins }) }, [...])
  useEffect(() => { setSettingsPlugins(settingsPlugins) }, [...])
  useEffect(() => { setStoreAutoUpdateNextAt(autoUpdateNextAt) }, [...])

  Since useEffect fires after paint, AppContent renders once per update with one-commit-stale store data. When a probe result arrives while the user is on a plugin detail page:
  1. App.tsx hooks recompute displayPlugins with new state
  2. But the store still has old displayPlugins
  3. AppContent renders, find() returns undefined, selectedPlugin is null
  4. ProviderDetailPage briefly flashes "Provider not found" for one frame

  Fix: Switch the three mirror effects to useLayoutEffect so the store is updated before paint. Or consolidate into a single atomic update:

  useLayoutEffect(() => {
    useAppDerivedStore.setState({ displayPlugins, navPlugins, settingsPlugins, autoUpdateNextAt })
  }, [displayPlugins, navPlugins, settingsPlugins, autoUpdateNextAt])

  This is safe since the app is Tauri-only (no SSR useLayoutEffect warnings).

  2. [SHOULD-FIX] selectedPlugin computed in useAppPluginViews is discarded

  Files: src/hooks/app/use-app-plugin-views.ts:68-77, src/App.tsx:198

  useAppPluginViews computes selectedPlugin from fresh hook data, but App.tsx only destructures { displayPlugins, navPlugins } — the value is silently dropped. AppContent then independently re-derives it from the stale derived store. This
  is both wasted computation and compounds issue #1.

  Fix: Either mirror selectedPlugin into the derived store alongside the other plugin views, or remove the dead computation from useAppPluginViews.

  3. [NICE-TO-HAVE] Defensive try/catch in tray updater

  File: src/hooks/app/use-tray-icon.ts:77

  If any synchronous code after trayUpdatePendingRef.current = true throws before reaching a .finally(finalizeUpdate) path, the pending flag stays stuck forever and no further tray updates fire. While current code paths are unlikely to
  throw synchronously, wrapping the timer callback body in try/catch with a finalizeUpdate() in the catch would be defensive hardening.

  4. [INFO] setTrayReady(true) ordering change

  File: src/hooks/app/use-tray-icon.ts:155-157

  The old code set trayReady before resolving the gauge icon resource. The new code sets it after. This delays the initial tray paint slightly if resolveResource is slow. Not a bug — the cancelled guard is correct — but worth noting if
  tray startup latency matters. The probe-triggered tray update during this narrow init window self-heals on the next update.

  ---
  Verified as Non-Issues

  - startBatch stability: handleProbeResult depends on [getErrorMessage, onProbeResult], both useCallback([], []) — stable. No listener churn in useProbeEvents.
  - Tray queuing infinite loop: The recursive scheduleTrayIconUpdate("probe", 0) in finalizeUpdate properly resets trayUpdateQueuedRef before recursing. Safe.
  - Cross-hook ref sharing: pluginStatesRef and manualRefreshIdsRef are useRef objects passed between sub-hooks within useProbe. They cannot detach.
  - useSettingsBootstrap effect stability: All 14 deps are Zustand store setters (stable) or useCallback([]) (stable). No re-run risk.
  - Stale closure in useSettingsTrayActions: trayShowPercentage is in the deps array and only changes via sequential user interaction. No practical race.
  - usePanel ResizeObserver churn: The [activeView, displayPlugins] deps cause teardown/recreate on every probe result, but this matches the old code exactly. Not a regression.
  - Fallback-to-home guard: The new pluginsMeta.some(...) check is an improvement — prevents spurious redirects before plugins load.

  ---
  Positives

  - Clean single-responsibility hooks with clear inputs/outputs
  - Tray queuing properly handles overlapping renders (old code silently dropped updates)
  - Desktop-only isTauri() guards correctly gate panel resize, escape-key, and event listeners
  - Good new test coverage for useAppPluginViews, useSettingsPluginList, useAppVersion, and app-derived-store
  - Zustand store design eliminates the fragile ref-sync pattern from the old code

  Minor Nits

  - handleBatchComplete in use-probe.ts:28 is an empty useCallback(() => {}, []) — remove if useProbeEvents can accept undefined
  - docs/app-state-architecture.md references "PR #209" directly — will go stale
  - ScheduleTrayIconUpdate type is duplicated in 3 files — extract to shared types

davidarny:
Addressed in `804da7e`.

Fixed both should-fix points from the review comment:

- Removed effect-based derived-state mirroring from `App` and passed `navPlugins`, `displayPlugins`, `settingsPlugins`, `autoUpdateNextAt`, and `selectedPlugin` directly into `AppShell`/`AppContent`.
- Added a loading guard in `useAppPluginViews` so fallback-to-`home` does not run while `pluginSettings` is still `null`.
- Added regression coverage for the loading case in `use-app-plugin-views.test.ts`.

Validation run:
- `bun run vitest src/hooks/app/use-app-plugin-views.test.ts src/App.test.tsx` (75 passing)
- `bun run build`

validatedev (approved):
Looks clean! LGTM!

PR #263, opened 2026-03-04
fix(amp): show credits in overview for credits-only users

davidarny:
## Summary
- Move Amp `Credits` line to `overview` scope so credits-only accounts show data on the overview card.
- Add a regression test for Amp's current credits-only response text variant with the automatic top-up hint.
- Update Amp provider docs to reflect the overview visibility behavior.

### Before
CleanShot 2026-03-04 at 09 46 21@2x

### After
CleanShot 2026-03-04 at 13 37 18@2x

## Testing
- `bunx vitest run plugins/amp/plugin.test.js`
- `bunx vitest run src/pages/overview.test.tsx`
- `bun run test:coverage` (fails on current `main` baseline too: branch coverage `89.61%` vs threshold `90%`)


davidarny:
@codex review

davidarny:
augment review

validatedev:
I was the one who implemented Amp. I thought I handled the credits thing correctly, but it seems I was wrong. Thank you 🥰

PR #268, opened 2026-03-06
fix(factory): prefer auth.encrypted over auth.json

sudoanmol:
## Problem

When a user re-authenticates with `droid`, it writes fresh tokens to `~/.factory/auth.encrypted`. However, the plugin reads `AUTH_PATHS` in order — `auth.json` first — so if a stale `auth.json` exists from a previous session, the plugin picks it up instead of the newer `auth.encrypted`, causing the refresh to fail with "Session expired".

## Fix

Swap the order so `auth.encrypted` (the current format, per docs) is checked before `auth.json` (the legacy format). Users on older droid versions that only have `auth.json` still work via fallback.

## Testing

Confirmed locally: had a stale `auth.json` (expired Feb 26) alongside a valid `auth.encrypted` (expires Mar 11). Before the fix the plugin showed "Session expired". After the fix it correctly reads `auth.encrypted` and loads usage data.

---
## Summary by cubic
Prefer ~/.factory/auth.encrypted over ~/.factory/auth.json when loading credentials, so re-authenticated sessions are honored and stale auth.json files no longer cause "Session expired" errors; legacy auth.json remains a fallback. Adds a regression test to confirm the preference and that no refresh is attempted when the token is valid.

<sup>Written for commit a6c3e3066f2f349fdd3db3bf21b1ccf9e84d718d. Summary will update on new commits.</sup>

### DIFF
--- a/plugins/factory/plugin.test.js
+++ b/plugins/factory/plugin.test.js
@@ -65,6 +65,42 @@ describe("factory plugin", () => {
     expect(result.lines.find((line) => line.label === "Standard")).toBeTruthy()
   })
 
+  it("prefers auth.encrypted over stale auth.json when both exist", async () => {
+    const ctx = makeCtx()
+    const pastExp = Math.floor(Date.now() / 1000) - 1000
+    const futureExp = Math.floor(Date.now() / 1000) + 7 * 24 * 60 * 60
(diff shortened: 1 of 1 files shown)


davidarny:
@codex review

PR #298, opened 2026-03-16
fix: load factory auth from droid v2 store

davidarny:
## Summary
- load Factory auth from Droid's current `auth.v2.file` + `auth.v2.key` store before falling back to legacy auth sources
- add host-side AES-256-GCM decrypt support so the Factory plugin can read the v2 encrypted auth payload
- keep the proactive refresh fallback fix and add regression coverage around v2 auth loading
- include the currently committed Rust/UI formatting changes that were already in the worktree

## Testing
- `bun run test --run plugins/factory/plugin.test.js`
- `cargo test --manifest-path src-tauri/Cargo.toml plugin_engine::host_api::tests`
- `bun run test:coverage` *(fails: global branch coverage is `89.58%`, below the `90%` threshold; PR created anyway per explicit user instruction)*

### DIFF
--- a/src-tauri/src/plugin_engine/host_api.rs
+++ b/src-tauri/src/plugin_engine/host_api.rs
@@ -2135,6 +2135,14 @@ mod tests {
         )
     }
 
+    fn node_generated_aes_256_gcm_vector_for_test() -> (&'static str, &'static str, &'static str) {
+        (
+            "CwsLCwsLCwsLCwsLCwsLCwsLCwsLCwsLCwsLCwsLCws=",
+            "BwcHBwcHBwcHBwcHBwcHBw==:yFbCs4LOJ0aj9NPNf5pfVA==:7PKjtOdATLClvaWrMw0b0M8Nov4KPhxwQX4hdczqQlcZi9Zhi6DjAoK+WolvMwuhPIk=",
+            r#"{"access_token":"token","refresh_token":"refresh"}"#,
+        )
+    }
+
     #[test]
     fn last_non_empty_trimmed_line_uses_final_value_when_stdout_is_noisy() {
         let stdout = "banner line\nanother message\n  sk-test-key-12345  \n";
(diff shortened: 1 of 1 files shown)


robinebers (commented):
Hey! This is Rob's AI reviewer. Thanks for the contribution!

Nice improvement overall. Loading the newer auth.v2 store first is the right direction, and the added test coverage helped a lot. Two things to fix before merge:

1. file-v2 refreshes are not persisted. refreshToken() updates the in-memory access_token and refresh_token, but saveAuth() skips source === \"file-v2\". Since the refresh response can return a new refresh_token, the next probe or app restart can fall back to stale auth from disk and fail once the old access token expires.

2. This PR adds a new shared plugin API helper, ctx.host.crypto.decryptAes256Gcm(), but docs/plugins/api.md was not updated. Since this changes the plugin API surface, the docs should move with the code.

Everything else I spot-checked here looks good, and the focused JS/Rust tests passed.

validatedev:
@codex review

validatedev:
I can approve this after you resolve @robinebers's comment (if relevant). I don't use droid so I can't test this though.

davidarny:
Addressing the AI review feedback:

- persisted refreshed Factory `file-v2` auth back to `~/.factory/auth.v2.file` using the same AES-256-GCM envelope/key path, so rotated `refresh_token` values survive the next probe/app restart
- documented the shared `ctx.host.crypto.decryptAes256Gcm(...)` and `ctx.host.crypto.encryptAes256Gcm(...)` helpers in `docs/plugins/api.md`

Verification:
- `bun run test --run plugins/factory/plugin.test.js`
- `cargo test --manifest-path src-tauri/Cargo.toml plugin_engine::host_api::tests`

davidarny:
Addressing the two Copilot review comments:

- replaced the direct `openssl` dependency with stable RustCrypto `aes-gcm 0.10.3`, which avoids native OpenSSL linkage/toolchain risk while keeping Droid's 16-byte IV envelope format
- added explicit validation for AES-256 key length, IV length, and GCM auth-tag length before decrypt/encrypt, so bad envelopes fail with clear errors instead of lower-level crypto init failures

Verification:
- `cargo test --manifest-path src-tauri/Cargo.toml plugin_engine::host_api::tests`

robinebers:
gonna fix and merge this @davidarny - cheers bud

Issue #106, opened 2026-02-07
all providers tab compact mode


Guria:
### Problem

As growing number of supported providers long vertical scroll appears on default all providers view.

### Proposed Solution

It might make sense to introduce compact view toggle on all providers tab

robinebers:
What do you have in mind? We have #47 already. Would love to come up with something.

Guria:
I would start with smaller spacings; progressbar height; collocate progressbar with label (maybe)

robinebers:
Could be alleviated by #301

PR #342, opened 2026-04-04
fix: keep tray panel height stable when provider content changes

prayzey:
## Description

Fixes the tray panel resizing vertically when switching between providers or views with different amounts of content.

Before this change, the window height followed the current content height, so short states and longer states made the whole panel jump up and down.

This change keeps a stable preferred panel height and lets the content scroll inside the panel when needed.

## Related Issue

N/A

Related to the general sizing/compactness discussion in #106, but this specifically fixes panel height jitter rather than adding a compact mode.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] New provider plugin
- [ ] Documentation
- [ ] Performance improvement
- [ ] Other (describe below)

## Testing

- [x] I added regression tests for stable panel height and small-screen clamping

## Screenshots

- Before: panel height changes depending on which provider state is visible
image

image

- After: panel height stays stable and only the inner content scrolls

image

newosu

## Checklist

- [x] I read `CONTRIBUTING.md`

---
## Summary by cubic
Keeps the tray panel height stable across provider/view changes so the window stops jumping. Uses a preferred 560px height with in-panel scrolling, clamps for small monitors, and refines the scrollbar and layout for better spacing.

- **Bug Fixes**
  - Stable `panelHeightPx` with a fixed card height in `AppShell`; dedup window resizes; recalc on DPI/scale changes and window moves.
  - Clamp height to monitor bounds with a small minimum; content scrolls inside the panel.
  - Thin `.panel-scroll` on the outer edge; padding moved to an inner content wrapper; footer now has `px-3` to match the gutter; updated styles and tests.

<sup>Written for commit 4410fb492a13bcdb8cd9b080541a2d66115b0147. Summary will update on new commits.</sup>

### DIFF
--- a/src/App.test.tsx
+++ b/src/App.test.tsx
@@ -8,6 +8,8 @@ const state = vi.hoisted(() => ({
   isTauriMock: vi.fn(() => false),
   trackMock: vi.fn(),
   setSizeMock: vi.fn(),
+  windowOnMovedMock: vi.fn(),
+  windowOnScaleChangedMock: vi.fn(),
   currentMonitorMock: vi.fn(),
   startBatchMock: vi.fn(),
   savePluginSettingsMock: vi.fn(),
@@ -172,7 +174,11 @@ vi.mock("@tauri-apps/api/path", () => ({
 }))
 
 vi.mock("@tauri-apps/api/window", () => ({
-  getCurrentWindow: () => ({ setSize: state.setSizeMock }),
+  getCurrentWindow: () => ({
+    setSize: state.setSizeMock,
+    onMoved: state.windowOnMovedMock,
+    onScaleChanged: state.windowOnScaleChangedMock,
+  }),
   PhysicalSize: class {
     width: number
     height: number
@@ -259,6 +265,8 @@ describe("App", () => {
     state.isTauriMock.mockReturnValue(false)
     state.trackMock.mockReset()
     state.setSizeMock.mockReset()
+    state.windowOnMovedMock.mockReset()
+    state.windowOnScaleChangedMock.mockReset()
     state.currentMonitorMock.mockReset()
     state.startBatchMock.mockReset()
     state.savePluginSettingsMock.mockReset()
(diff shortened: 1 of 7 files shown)


validatedev:
@codex review

validatedev (changes requested):
One issue I have seen behaviorally that users expect a scrollbar to appear while scrolling, similar to how browsers handle it, and I don't have such thing while testing your build. Not sure whether it should stay visible at all times or only show during active scrolling. Nice thinking though, thanks!

validatedev:
CleanShot 2026-04-05 at 01 51 21@2x

Thanks, but the scrollbar is not in the correct place.

prayzey:
Screenshot 2026-04-05 at 12 00 16 AM

@validatedev How is this ? moved it further away from the content. Unless you want it in a different area. Thanks

validatedev:
CleanShot 2026-04-05 at 01 59 35@2x
Thanks again for the quick fix and the first iteration of the implementation, but I still feel like it shouldn't look like that. As you can see, there's no distance between the scroll bar and the content, and it does not appeal to me. Don't think that I'm being unnecessarily picky, I am trying to make sure the app looks right. The scroll bar on other apps tends to be in the rightmost area. I feel like we should have something like that instead:
CleanShot 2026-04-05 at 02 12 31@2x

prayzey:
Thanks for the feedback! How does this feel?

image

Newscroll

validatedev:
Hmm, now we have a gap on the right even without any scrollbar, which doesn’t feel ideal. But thanks, I appreciate your work. Let me work on it as well.

validatedev:
CleanShot 2026-04-05 at 14 34 51@2x

I'm still working on it but I have eliminated all of the padding problems.

davidarny:
@prayzey @validatedev guys make sure to support different macos scrollbar modes. maybe that's a root of your issues?

CleanShot 2026-04-05 at 23 52 16@2x

However not sure if it works the same way in tauri app as it does in browser. So basically in first two options scrollbar is overlayed, in last option a space for scrollbar will be added.

https://github.com/user-attachments/assets/ebe4f669-3e49-4bc8-bd11-905194668b64

prayzey:
Thanks for pointing that out. 
Yes it does actually change the behaviour a bit.   

How would you suggest we move forward so that the design can be consistent? Thank you. 
image

image

robinebers:
So this is actually a feature, not a bug, and it's intended for you to see all your providers at all times, even if you have multiple. Personally, I would hate if I suddenly have to scroll, and I think most people won't use more than 2-3.

prayzey:
Thanks. That makes sense. I'll close the PR now.

PR #344, opened 2026-04-05
Filter control for what models on main tab

mitchcapper:
## Description

This allows you to exclude certain models from the main tab in order to help make it more concise.    You can still see all models on the individual provider's tab.

## Type of Change

- [ ] Bug fix
- [x ] New feature
- [ ] New provider plugin
- [ ] Documentation
- [ ] Performance improvement
- [ ] Other (describe below)

## Testing

- [X] I ran `bun run build` and it succeeded
- [X] I ran `bun run test` and all tests pass
- [-] I tested the change locally with `bun tauri dev`
I am running windows so could only test with the https://github.com/Rana-Faraz/usage-tray-windows fork

## Screenshots

The main with filtered models:
image

The settings to pick what models (only shows the model selection when the provider is enabled):
image

## Checklist

- [X] I read [CONTRIBUTING.md](../CONTRIBUTING.md)
- [X] My PR targets the `main` branch
- [X] I did not introduce new dependencies without justification

---
## Summary by cubic
Add per-provider controls to hide specific model lines on the Overview tab, making the main tab more concise. Settings are saved and do not affect provider-specific tabs.

- **New Features**
  - In Settings, enabled providers show checkboxes for each `scope: "overview"` line to include/exclude them on Overview.
  - Added `disabledOverviewLabels` to plugin settings with persistence, normalization, and equality checks.
  - Overview filters out disabled labels in `ProviderCard`; provider tabs remain unchanged. Wiring added via `onToggleOverviewLabel` through `App`, `AppContent`, and hooks.

<sup>Written for commit 7361b98f76a47dcaab2158aa4eb8953542653afa. Summary will update on new commits.</sup>

### DIFF
--- a/src/App.tsx
+++ b/src/App.tsx
@@ -158,6 +158,7 @@ function App() {
   const {
     handleReorder,
     handleToggle,
+    handleToggleOverviewLabel,
   } = useSettingsPluginActions({
     pluginSettings,
     setPluginSettings,
@@ -241,6 +242,7 @@ function App() {
         onRetryPlugin: handleRetryPlugin,
         onReorder: handleReorder,
         onToggle: handleToggle,
+        onToggleOverviewLabel: handleToggleOverviewLabel,
         onAutoUpdateIntervalChange: handleAutoUpdateIntervalChange,
         onThemeModeChange: handleThemeModeChange,
         onDisplayModeChange: handleDisplayModeChange,
--- a/src/components/app/app-content.tsx
+++ b/src/components/app/app-content.tsx
@@ -26,6 +26,7 @@ export type AppContentActionProps = {
   onRetryPlugin: (id: string) => void
   onReorder: (orderedIds: string[]) => void
   onToggle: (id: string) => void
+  onToggleOverviewLabel: (pluginId: string, label: string) => void
   onAutoUpdateIntervalChange: (value: AutoUpdateIntervalMinutes) => void
   onThemeModeChange: (mode: ThemeMode) => void
   onDisplayModeChange: (mode: DisplayMode) => void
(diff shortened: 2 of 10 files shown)


validatedev:
@codex review

davidarny:
@mitchcapper thanks for the PR, but I doubt on this one.

Is it something you really need? What should happen when you check something that disappears later? What should happen when you uncheck all? Does it even make sense?

All these edge cases make me thing that this kind of customization will bring much more complexity then being usefull.

mitchcapper:
I am not sure about other people but I check my usage pretty often this consolidating it into one place is great.   The problem becomes when you have many providers in use at the same time.

Not only is there visual clutter showing models users may not care about (ie flash or chat) but it also means you have to scroll to see what could otherwise be presented on all one screen. 

"What should happen when you check something that disappears later":   Everything is checked by default, unchecking a model adds it to a "hide" list.   This also means new models show up by default.   In the model goes away it disappears for everyone so the fact you had it checked would not seem to matter.

"What should happen when you uncheck all?"   Nothing would show.  It would kind of be like asking what would happen if you removed all the items from the start menu or remove all your items from your favorites list, it would be empty.  Should you ever uncheck all of a specific provider?  Probably not i would just move them to the bottom, although I could see some people preferring the cleaner look of them not on the main page at all potentially.    It would be trivial to make it recheck everything if you uncheck it all, or you could not uncheck the last item, but again I am not really sure that is a condition one needs to account for.

This is a high/frequent touch tool the more streamlined and showing just what the user wants the better I figure.

If the feature isn't needed its no problem, I can maintain my own patch (or see about just having it merged downstream).  I just try to contribute back things that seem useful to others.

If you want to scope down the footprint one could make this a json only feature so no UI for editing, if your concern is UI clutter one could have a json flag controlling if it was enabled or not.

robinebers:
Like David said, I think the idea is good. It just needs a little bit more thinking through, and a lot of people probably don't have this problem, to be honest. Closing this for now, but thank you for your PR!

Issue #377, opened 2026-04-13
Claude plugin: /api/oauth/usage returns 429 due to too-frequent polling


zergzorg:
### Problem

Environment:
  - OpenUsage v0.6.14
  - macOS
  - Proxy: SOCKS5

  Problem:

  The Claude plugin polls https://api.anthropic.com/api/oauth/usage too frequently (~every 10 seconds), causing
  Anthropic's API to return HTTP 429 rate limit errors. As a result, Claude usage stats are never successfully fetched.

  Log evidence:
  [12:13:15] [plugin:claude] HTTP GET https://api.anthropic.com/api/oauth/usage
  [12:13:16] [plugin:claude] HTTP GET https://api.anthropic.com/api/oauth/usage -> 429
             {"error": {"type": "rate_limit_error", "message": "Rate limited. Please try again later."}}
  [12:13:24] [plugin:claude] HTTP GET https://api.anthropic.com/api/oauth/usage  ← повтор через ~10 сек
  [12:13:25] [plugin:claude] HTTP GET https://api.anthropic.com/api/oauth/usage -> 429  ← снова 429

  Note: credentials are read from keychain successfully — the issue is not authentication but polling frequency.

  Expected behavior:
  - Polling interval should be increased (e.g. 5–10 minutes minimum)
  - On 429 response, implement exponential backoff before retrying
  - Ideally: respect Retry-After header if Anthropic returns one

  Contrast: The Codex plugin polls the same batch but fetches successfully — likely because ChatGPT's endpoint has a
  higher rate limit or the plugin uses a longer interval.

### Proposed Solution

  Implement exponential backoff with a minimum polling interval of 5 minutes for the Claude plugin. On a 429 response,
  pause polling and double the retry interval (e.g. 5m → 10m → 20m, capped at 1 hour). If the response includes a
  Retry-After header, respect it instead of guessing. Additionally, expose a configurable polling interval in settings
  so users can tune it manually.

zergzorg:
Impact:

  Aggressive polling by OpenUsage exhausts the rate limit on /api/oauth/usage, causing Claude Code's own built-in /usage
   command to also fail with the same 429 error — making it impossible to check usage from the CLI while OpenUsage is
  running.

garanda21:
Duplicate #343 

The solution proposed here https://github.com/robinebers/openusage/issues/343#issuecomment-4198941997 works also for me. `claude /login` then refresh provider on OpenUsage

PR #378, opened 2026-04-13
fix(claude): graceful 429 rate limit handling with Retry-After support

### LINKED ISSUE #376: Refresh now button
### Problem

We would like to have a straightforward button to refresh "now." Currently, when you click on options, there is a "reload now" feature; however, we prefer if this could be instantaneous.

### Proposed Solution

having a button refresh now on all tabs, and one on the main to refresh all ai

### Alternatives Considered

_No response_

### Additional Context

_No response_

zergzorg:
## Summary

Closes #376

- Parse `Retry-After` header on 429 (supports both seconds and HTTP-date formats)
- Show amber **"Rate limited, retry in ~Xm"** badge instead of throwing an error
- ccusage data (Today / Yesterday / Last 30 Days) continues to display when rate limited
- Add a Note line explaining that live usage data may be stale
- Removed a fake retry loop that hammered the API 3× immediately on 429 with no actual delay (made rate limiting worse)

## Changes

**`plugin.js`**
- `parseRetryAfterSeconds(headers)` — parses Retry-After header
- `fetchUsageWithRetryAfter(ctx, accessToken)` — single request, attaches `_retryAfterSeconds` on 429
- `probe()` — on 429: sets `rateLimited = true`, shows Status badge + Note, skips throwing

**`plugin.test.js`** — 6 new tests covering:
- 429 shows badge without throwing
- Retry-After seconds parsed correctly
- Fallback message when Retry-After missing
- HTTP-date Retry-After format
- Rate limited status with ccusage data still present
- Generic "try again later" message

## Test plan

- [ ] Verify badge appears with correct wait time when API returns 429 + Retry-After
- [ ] Verify ccusage rows (Today/Yesterday/Last 30 Days) still render during rate limit
- [ ] Verify no error thrown on 429
- [ ] Verify normal flow unaffected when API returns 200

---
## Summary by cubic
Gracefully handle Claude API 429s by honoring Retry-After, backing off between probes, and showing a wait badge instead of throwing. Adds a 5‑minute minimum usage fetch interval with a bypass right after rate‑limit windows and caches the last usage so data stays visible.

- **Bug Fixes**
  - Parse `Retry-After` as seconds or HTTP‑date; allow `0` (“retry now”); use `Math.ceil` and strict null checks; parse immediately after the 429 check.
  - On 429: don’t throw; record a backoff (`Retry-After` or 5m default); skip API calls until it expires; show amber “Rate limited, retry in ~Xm/~now” badge + Note; keep cached usage/plan data.
  - Polling guard: never hit the usage API more than once per 5 minutes; reuse cached response when the interval hasn’t elapsed; bypass the min‑interval right after a rate limit so short `Retry-After` windows aren’t swallowed.
  - Tests: add `_resetState`, use fake timers, stub `ctx.util.requestJson`, and cover seconds/HTTP‑date/`0`, missing header, no‑calls‑during‑backoff, resume‑after‑expiry with bypass, min‑interval skip, cached‑data during limits; tighten the resume‑after‑rate‑limit assertion to check for the amber badge correctly.
  - Settings: stop Base UI `Checkbox` click bubbling and restore `onCheckedChange` so each toggle saves exactly once; re‑query the checkbox between clicks and use pre‑normalised settings in tests to avoid an init save.

<sup>Written for commit a80dd86813e7ae7e4a1deb240bc9875315bc1b86. Summary will update on new commits.</sup>


zergzorg:
@augmentcode[bot] @cubic-dev-ai please re-review — issues #1 (Retry-After:0), #2 (Math.round), #3 (flaky test) are resolved in commits `7f9c0a9` and `2877eb2`. Please dismiss stale review comments.

zergzorg:
@robinebers @validatedev @davidarny

CI blocks merge, but this is a **pre-existing** issue — not from this PR.

**App.test.tsx:785**: `expected savePluginSettings to be called 2 times, but got 3 times` — flaky mock count assertion.

Evidence:
- CI on **main** also fails (run 24339426410 — 'feat: add Fireworks AI plugin')
- cubic-dev-ai re-review: SUCCESS
- All 7 rate-limiting tests in this PR: PASS

@robinebers safe to merge (CI unrelated) or fix the flaky test separately.

validatedev:
@codex review

zergzorg:
## CI failure is pre-existing on `main`

The failing test (`src/App.test.tsx > App > toggles plugins in settings`) is **not caused by this PR**. The same test fails on `main` with the same error in the Fireworks AI plugin PR merged before this one — see run [24339426410](https://github.com/robinebers/openusage/actions/runs/24339426410).

My changes only touch `plugins/claude/plugin.js` and `plugins/claude/plugin.test.js`.

---

Also addressed the bot review comments in the latest commit:
- Removed `fetchUsageWithRetryAfter` helper that mutated the HTTP response object — `parseRetryAfterSeconds` is now called directly in the 429 handler
- Changed `||` → `??` when reading the `Retry-After` header (safe for `0` values)
- Made the HTTP-date test deterministic with `vi.useFakeTimers()` + a fixed future timestamp

zergzorg:
## CI is now green 🟢

Turns out the `toggles plugins in settings` failure had **three compounding causes**, all pre-existing before this PR. Here's the full post-mortem in case it's useful:

---

### Root cause 1 — `settings.tsx` double-fire (introduced in `57cc5bd`)

Commit `57cc5bd` ("clickable provider rows") moved the toggle from `Checkbox.onCheckedChange` to `div.onClick` so the entire row is clickable. The problem: **Base UI's `CheckboxRoot` internally dispatches a synthetic `PointerEvent('click')` on its hidden `<input>` element** after every user click (to keep the native input in sync). Both the visible `<span>` click *and* the hidden `<input>` click bubbled up to the row `<div onClick=onToggle>`, firing `onToggle` — and therefore `savePluginSettings` — **twice per click**.

**Fix:** wrap `<Checkbox>` in a `<span onClick={(e) => e.stopPropagation()}>` so neither event reaches the row div. Restore `onCheckedChange` on the Checkbox to keep exactly one call per click. (`src/pages/settings.tsx`)

---

### Root cause 2 — init normalisation save not accounted for in test

`beforeEach` seeds `loadPluginSettings` with `{ order: ["a"], disabled: [] }`. On startup, `normalizePluginSettings` appends plugin `"b"` to `order` **and** to `disabled` (since `"b"` is not in `DEFAULT_ENABLED_PLUGINS`), producing `{ order: ["a","b"], disabled: ["b"] }`. Because the stored and normalised forms differ, `savePluginSettings` fires once during bootstrap — before the test even clicks anything.

**Fix:** supply the already-normalised form `{ order: ["a","b"], disabled: ["b"] }` directly in this test so `arePluginSettingsEqual` returns `true` and no init save fires. (`src/App.test.tsx`)

---

### Root cause 3 — stale DOM reference after Checkbox remount

`<Checkbox key={${plugin.id}-${plugin.enabled}} .../>` changes its `key` on every toggle, causing React to unmount the old node and mount a fresh one. The test was caching the checkbox reference before the first click, so the second `userEvent.click(pluginCheckbox)` was operating on a detached (stale) DOM element — effectively a no-op.

**Fix:** re-query with `screen.findAllByRole("checkbox")` before *each* click so we always hold a live reference. (`src/App.test.tsx`)

zergzorg:
## Summary of changes

This PR fixes the root cause described in issue #377 — the Claude plugin was polling the Anthropic usage API too frequently (~every 10 s), consistently hitting the 429 rate limit on every refresh cycle.

### Root cause fix

Added module-scope rate-limit state that persists across `probe()` calls:

- **`MIN_USAGE_FETCH_INTERVAL_MS` (5 min)** — enforces a minimum interval between usage API calls, matching the reset cadence of the usage data itself. Intermediate `probe()` calls reuse the last cached response silently.
- **`rateLimitedUntilMs`** — when a 429 is received, the next API call is blocked until the `Retry-After` window expires. If no `Retry-After` header is present, a 5-minute default backoff is applied.
- **`cachedUsageData`** — last successful response is cached and shown to the user even while rate-limited, so the UI stays populated.

A key edge case is handled: when `Retry-After` is shorter than the 5-minute poll throttle (e.g. `Retry-After: 60`), the min-interval guard is bypassed on recovery so the plugin actually retries after the specified window instead of waiting the full 5 minutes.

### What was NOT the fix

Showing a "Rate limited" badge in the UI (the previous approach in earlier commits) was only masking the symptom. The plugin was still hammering the API on every tick — it just wasn't surfacing the error. The fix prevents the redundant requests from being made in the first place.

Closes #377

robinebers (approved):
LGTM and thank you

PR #386, opened 2026-04-16
feat: preserve usage data during refresh (stale-while-revalidate)

### LINKED ISSUE #385: Preserve current values during refresh (stale-while-revalidate) + show "last updated" timestamp
### Problem

When a provider is refreshed — via manual "Refresh all", per-card retry, or scheduled auto-update — the currently displayed usage values (numbers, progress bars, reset times) vanish and are replaced by skeleton placeholders until the new fetch finishes. On a slow provider this means a user who opens the panel just to glance at their remaining usage sees no data at all for the duration of the refresh, even though perfectly good data was on screen a moment earlier.

On top of that, there is no "last updated" indicator anywhere on the card, so you can't tell whether what you're looking at is 10 seconds old or 10 minutes old.

Root cause in the current code:

[src/hooks/app/use-probe-state.ts](https://github.com/DoozyX/openusage/blob/claude/preserve-usage-on-refresh-1pNCN/src/hooks/app/use-probe-state.ts) — setLoadingForPlugins() explicitly sets data: null when a refresh starts:
next[id] = {
  data: null,       // ← wipes existing values
  loading: true,
  error: null,
  lastManualRefreshAt: existing?.lastManualRefreshAt ?? null,
}
[src/components/provider-card.tsx](https://github.com/DoozyX/openusage/blob/claude/preserve-usage-on-refresh-1pNCN/src/components/provider-card.tsx) then branches on loading: if true it renders <SkeletonLines />, otherwise it renders the real lines. It's a hard swap, not an overlay.
PluginState ([src/hooks/app/types.ts](https://github.com/DoozyX/openusage/blob/claude/preserve-usage-on-refresh-1pNCN/src/hooks/app/types.ts)) has lastManualRefreshAt, but it's only used for the retry cooldown — it isn't shown to the user as a freshness timestamp.
This was originally introduced in #5 ("Loading state while just after opening / while updating"), which was the right fix for cold start (no data yet). The regression is that the same skeleton pattern is also applied to subsequent refreshes where we already have perfectly valid cached data.

### Proposed Solution

Adopt a stale-while-revalidate pattern for refreshes:

Don't clear data on refresh. Only set loading: true. Keep the previously fetched PluginOutput visible. Clear it only on a "hard reset" (e.g. provider disabled → re-enabled, auth changed) or if the new fetch errors and there's no cached data.
Render a subtle overlay indicator on the progress bar while loading && data != null:
a small shimmer/pulse on the progress bar track, or
a tiny spinner badge next to the provider title, or
an indeterminate bar overlaid on the existing determinate bar.
The existing skeleton behavior stays as-is for the first load (when data == null).
3. Add a lastUpdatedAt: number | null field to PluginState, set from the probe:result event timestamp in use-probe-events.ts. Display it on the card as a relative string ("Updated 2m ago", "just now"), auto-ticking every ~30s. Make it an accessible <time> element with an absolute title tooltip.
4. On error during refresh, keep the last-known values visible but decorate the card with a small warning icon + tooltip showing the error. Don't replace the numbers with an error state if we have stale data to show — this is how Activity Monitor, iStat Menus and most system-status UIs handle transient failures.

Why this matches established best practices
TanStack Query — placeholderData: keepPreviousData is the canonical pattern for "show old data while new data loads" and is explicitly recommended for exactly this UX ([docs](https://tanstack.com/query/latest/docs/framework/react/guides/placeholder-query-data)).
SWR (the library that named the pattern) — default behavior returns stale cache immediately and revalidates in the background; isValidating is surfaced separately from isLoading so the UI can show a small indicator without swapping content. The HTTP Cache-Control: stale-while-revalidate directive ([RFC 5861](https://www.rfc-editor.org/rfc/rfc5861)) formalizes the same idea.
Apple Human Interface Guidelines — Progress indicators: "When possible, show progress without interrupting the content people are viewing." Menu-bar / status apps (Activity Monitor, iStat Menus, Stats) all update metrics in place and never blank their readouts during a poll.
Nielsen Norman Group on skeleton screens: skeletons are appropriate for first paint, but re-showing them on every refresh causes content flicker and is perceived as slower than simply leaving old content visible with a small activity hint.
Material Design — Progress indicators: a linear indeterminate indicator shown over/beside existing content is the recommended pattern for "the data on screen is being refreshed".
Acceptance criteria
not done
Clicking "Refresh all" or per-provider retry does not cause existing numbers, progress bars or reset times to disappear.
not done
A small, non-blocking activity indicator is visible on each card being refreshed (e.g. pulse on progress bar or spinner next to title).
not done
Each provider card shows a relative "Updated Xm ago" label, with a tooltip giving the absolute time.
not done
First-time load (no cached data) still shows the existing skeleton.
not done
A failed refresh while cached data exists keeps the data on screen and surfaces the error as a non-destructive warning, not a full error state.
not done
Auto-refresh on interval behaves identically (no flicker, overlay indicator only).
Notes for implementers
The minimal code-level change is in [src/hooks/app/use-probe-state.ts](https://github.com/DoozyX/openusage/blob/claude/preserve-usage-on-refresh-1pNCN/src/hooks/app/use-probe-state.ts) setLoadingForPlugins() — stop nulling data, preserve existing?.data instead. Then thread lastUpdatedAt through handleProbeResult() and render the overlay + timestamp in [src/components/provider-card.tsx](https://github.com/DoozyX/openusage/blob/claude/preserve-usage-on-refresh-1pNCN/src/components/provider-card.tsx). The batch/event plumbing in [src/hooks/use-probe-events.ts](https://github.com/DoozyX/openusage/blob/claude/preserve-usage-on-refresh-1pNCN/src/hooks/use-probe-events.ts) does not need to change.

### Alternatives Considered

_No response_

### Additional Context

_No response_

DoozyX:
Stop clearing provider data when a refresh starts. Instead of replacing content with skeleton loaders on every refresh, keep the existing values visible and show a subtle shimmer overlay on progress bars. This avoids the jarring flash-to-skeleton UX on subsequent refreshes while keeping skeletons for first-time loads when no data exists.

Changes:
- setLoadingForPlugins preserves existing data instead of nulling it
- setErrorForPlugins preserves stale data on transient failures
- handleProbeResult preserves stale data on error responses
- Added lastUpdatedAt field to PluginState, shown as relative timestamp
- Progress bar gains a shimmer overlay when refreshing with stale data
- Error during refresh shows inline warning instead of replacing content

Closes robinebers/openusage#385

https://claude.ai/code/session_01F1GBhgCyfdf36cqnq5Lurd

## Type of Change

- [ ] Bug fix
- [x] New feature
- [ ] New provider plugin
- [ ] Documentation
- [ ] Performance improvement
- [ ] Other (describe below)

## Testing

- [ ] I ran `bun run build` and it succeeded
- [ ] I ran `bun run test` and all tests pass
- [ ] I tested the change locally with `bun tauri dev`

## Checklist

---
## Summary by cubic
Preserves usage data during refresh to remove the flash-to-skeleton effect. Progress bars shimmer while refreshing; errors show inline with stale data, and an “Updated X ago” tooltip appears on the refresh icon.

- **New Features**
  - Keep existing provider data during refresh and transient errors; skeletons only on first load.
  - Add `lastUpdatedAt`; show “Updated X ago” only as a tooltip on the refresh button; clamp negative deltas to “just now”.
  - Add a shimmer overlay to progress bars when refreshing; respects reduced motion.
  - Show an inline warning with a tooltip when refresh fails and stale data exists; show the full error only when no data.
  - Remove periodic timestamp updates; tooltip computes freshness on open to avoid background timers.

<sup>Written for commit 7afc4fe71aa8e7217167614c10c4823bac2328a5. Summary will update on new commits.</sup>

### DIFF
--- a/src/components/app/app-content.test.tsx
+++ b/src/components/app/app-content.test.tsx
@@ -53,6 +53,7 @@ function createProps(): AppContentProps {
       loading: false,
       error: null,
       lastManualRefreshAt: null,
+      lastUpdatedAt: null,
     },
     onRetryPlugin: vi.fn(),
     onReorder: vi.fn(),
--- a/src/components/provider-card.test.tsx
+++ b/src/components/provider-card.test.tsx
@@ -815,6 +815,7 @@ describe("ProviderCard", () => {
         name="SWR"
         displayMode="used"
         loading
+        lastUpdatedAt={Date.now() - 60_000}
         lines={[
           { type: "text", label: "Label", value: "Value" },
           { type: "progress", label: "Session", used: 32, limit: 100, format: { kind: "percent" } },
@@ -830,6 +831,28 @@ describe("ProviderCard", () => {
     expect(screen.getByText("32%")).toBeInTheDocument()
     // Progress bar shows shimmer overlay
     expect(document.querySelector('[data-slot="progress-refreshing"]')).toBeTruthy()
+  })
+
+  it("skips skeleton on refresh when lastUpdatedAt set even if filtered lines are empty", () => {
+    render(
+      <ProviderCard
+        name="FilteredEmpty"
+        displayMode="used"
+        loading
+        lastUpdatedAt={Date.now() - 60_000}
+        scopeFilter="overview"
+        skeletonLines={[
+          { type: "progress", label: "Session", scope: "overview" },
+        ]}
+        lines={[
+          { type: "text", label: "DetailOnly", value: "Hidden" },
+        ]}
+      />
+    )
+    // skeleton label for "Session" should not render because we have prior data (lastUpdatedAt set)
+    expect(screen.queryByText("Session")).toBeNull()
+    // and the detail-scoped line stays filtered out
+    expect(screen.queryByText("DetailOnly")).toBeNull()
   })
 
   it("renders skeleton on first load when no stale data exists", () => {
@@ -851,6 +874,7 @@ describe("ProviderCard", () => {
         name="StaleErr"
         displayMode="used"
         error="Couldn't update data. Try again?"
+        lastUpdatedAt={Date.now() - 60_000}
         lines={[
           { type: "progress", label: "Session", used: 40, limit: 100, format: { kind: "percent" } },
         ]}
@@ -877,7 +901,7 @@ describe("ProviderCard", () => {
     expect(screen.getByText("Nope")).toBeInTheDocument()
   })
 
-  it("shows relative last-updated timestamp", () => {
+  it("shows relative last-updated timestamp in retry tooltip", () => {
     vi.useFakeTimers()
     const now = new Date("2026-02-02T00:05:00.000Z")
     vi.setSystemTime(now)
@@ -885,6 +909,7 @@ describe("ProviderCard", () => {
       <ProviderCard
         name="Updated"
         displayMode="used"
+        onRetry={() => {}}
         lastUpdatedAt={now.getTime() - 120_000}
         lines={[{ type: "text", label: "Label", value: "Value" }]}
       />
@@ -901,6 +926,7 @@ describe("ProviderCard", () => {
       <ProviderCard
         name="Fresh"
         displayMode="used"
+        onRetry={() => {}}
         lastUpdatedAt={now.getTime() - 5_000}
         lines={[{ type: "text", label: "Label", value: "Value" }]}
       />
@@ -917,12 +943,25 @@ describe("ProviderCard", () => {
       <ProviderCard
         name="Skew"
         displayMode="used"
+        onRetry={() => {}}
         lastUpdatedAt={now.getTime() + 60_000}
         lines={[{ type: "text", label: "Label", value: "Value" }]}
       />
     )
     expect(screen.getByText(/Updated just now/)).toBeInTheDocument()
     vi.useRealTimers()
+  })
+
+  it("omits retry tooltip content when lastUpdatedAt is null", () => {
+    render(
+      <ProviderCard
+        name="NoTimestamp"
+        displayMode="used"
+        onRetry={() => {}}
+        lines={[{ type: "text", label: "Label", value: "Value" }]}
+      />
+    )
+    expect(screen.queryByText(/Updated/)).toBeNull()
   })
 })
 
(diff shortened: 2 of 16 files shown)


robinebers:
Hi, thank you and solid change.

This was long on the todo, and the main reason we haven't done it yet was the UI implication.

Just having a little label there feels wasteful in an otherwise tight app.

We were considering showing 'stale' data with a lower opacity, or maybe showing a "Last update failed" error only if the probe cam back with an error.

Would love to explore something in that direection.

DoozyX:
Hey @robinebers — pushed another round. Addresses your feedback plus the rest of the cubic/Copilot comments.

**Your feedback:**
- Removed the always-visible "Updated X ago" footer. Felt wasteful in a tight app, agreed.
- "Updated Xm ago" is now shown only as a tooltip on the per-card refresh icon (hover the card → icon fades in → hover icon for the timestamp). Zero layout space, still discoverable, and showing refresh-related info on the refresh control reads naturally.
- Stale error path is unchanged: "Last update failed" inline warning only shown when a refresh errors and we have prior data.

**cubic (age-based ticker memo):**
- Dropped the ticker bucket entirely. The memo's deps were stale so long-lived cards kept the initial cadence. Tooltip content now reads `Date.now()` fresh on each open — no ticker needed for the timestamp at all.

**Copilot:**
- `hasStaleData` now uses `lastUpdatedAt != null || filteredLines.length > 0` — the true "has ever loaded" signal. Fixes the case where the scope filter hides all lines but we still shouldn't flash back to a skeleton.
- All `Boolean(lastUpdatedAt)` / `!lastUpdatedAt` truthy checks replaced with `!= null` / `== null` (handles timestamp `0` correctly).
- Added `@media (prefers-reduced-motion: reduce)` guard so the progress shimmer respects reduced-motion preferences.
- Added regression tests for the filtered-empty path and for the "no tooltip content when lastUpdatedAt is null" case.

Tested locally with \`bun tauri dev\` — live Claude + Codex probes worked end-to-end, data stays visible during refresh, shimmer shows, timestamp tooltip appears on refresh icon.

Would love another look.

robinebers:
@DoozyX Thank you, let's try this.

PR #419, opened 2026-04-29
fix(codex): lazy-load keychain auth fallback

validatedev:
## Description

Only read Codex Keychain auth when file auth is missing, invalid, or fails with an auth/session error. Preserve stale `auth.json` recovery while surfacing non-auth usage failures directly.

Fixes a Codex-only macOS Keychain prompt introduced after #413.

The Codex plugin now tries file auth first and only reads the `Codex Auth` Keychain fallback when file auth is missing, invalid, or fails with an auth/session error. This preserves stale `auth.json` recovery while avoiding Keychain prompts on every auto-refresh when file auth already works.

The auth fallback is intentionally driven by a small set of exact, user-facing auth error messages. The plugin treats those messages as its auth-failure contract: when file auth throws one of them, it may try the next auth source. Other errors remain loud and do not fall back.

Non-auth usage failures now surface directly instead of silently falling through to Keychain, including HTTP 500, invalid usage JSON, request exceptions, and API-key-only auth.

## Type of Change

- [x] Bug fix
- [x] Documentation

## Testing

## Checklist

### DIFF
--- a/docs/providers/codex.md
+++ b/docs/providers/codex.md
@@ -80,7 +80,7 @@ OpenUsage Codex plugin auth lookup order:
 3. `~/.codex/auth.json`
 4. macOS keychain service `Codex Auth` (fallback)
 
-If file OAuth auth is missing, invalid, or fails with an auth/session error during refresh or usage lookup, OpenUsage tries the macOS keychain fallback. Non-auth usage failures, such as server errors or invalid responses, are shown directly.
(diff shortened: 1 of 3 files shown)


validatedev:
@codex review

robinebers:
Lovely, thank you for fixing this. 

Pushing this out ASAP.

PR #424, opened 2026-05-01
fix(claude): support hashed macOS keychain service name (closes #423)

### LINKED ISSUE #423: Claude provider does not detect hashed Claude Code macOS Keychain credentials
### Description

OpenUsage 0.6.21 shows the Claude provider as expired even though Claude Code reports a successful login from the CLI.

The initial error shown in OpenUsage was:

```
Session expired. Run `claude` to log in again.
```

Following the instruction, I ran `claude`, logged out, and logged in again, but the OpenUsage Claude card still showed the same expired-session message.

This led me to inspect where OpenUsage is reading Claude credentials from. From the current OpenUsage Claude plugin behavior, it appears to check the legacy credential locations:

- `~/.claude/.credentials.json`
- macOS Keychain service `Claude Code-credentials`

On my machine, `~/.claude/.credentials.json` does not exist. I also found an old Keychain item named `Claude Code-credentials`.

After deleting that legacy Keychain item, I logged out and logged back into Claude Code. Claude Code reported a successful login. I expected the legacy Keychain service `Claude Code-credentials` to be recreated after a fresh login, but it was not. The OpenUsage error changed from ``Session expired. Run `claude` to log in again.`` to ``Not logged in. Run `claude` to authenticate.``

This suggests that OpenUsage was previously reading a stale legacy Keychain item, attempting to refresh it, and then reporting `Session expired`. Once that stale item was removed, OpenUsage had no compatible credential source left and fell back to `Not logged in`. A fresh Claude Code login did not recreate the legacy service, because the active credential appears to be stored under a hashed Keychain service instead:

```
Claude Code-credentials-<8-character-hash>
```

On my machine, the suffix matches:

```
sha256("/Users/<user>/.claude").slice(0, 8) == 8bab74ae
```

Current analysis of the OpenUsage source shows that the Claude plugin looks for `~/.claude/.credentials.json` or the Keychain item `Claude Code-credentials`. However, recent Claude Code versions appear to store macOS credentials under `Claude Code-credentials-<hash>` instead of the legacy service without a hash suffix. This also seems related to the upstream Claude Code Keychain / `CLAUDE_CONFIG_DIR` behavior discussed in:

- https://github.com/anthropics/claude-code/issues/19456#issuecomment-3967333078
- https://github.com/anthropics/claude-code/issues/20553#issuecomment-3967317782

So the likely issue is that newer Claude Code stores macOS credentials under:

```text
Claude Code-credentials-${sha256(expandedClaudeConfigDir).slice(0, 8)}
```

while OpenUsage still only checks the legacy service name:

```text
Claude Code-credentials
```

### Steps to Reproduce

First, confirm Claude Code is logged in:

```
claude auth login --claudeai
claude auth status --text
```

Claude Code reports a successful login.

Then confirm that the legacy credential locations used by OpenUsage are absent:

```
test ! -f ~/.claude/.credentials.json && echo "no legacy credentials file"

security find-generic-password -s "Claude Code-credentials" >/dev/null 2>&1 \
  || echo "no legacy Keychain service"
```

Now confirm that Claude Code has an active hashed Keychain credential instead:

```
HASH=$(CONFIG_DIR="${CLAUDE_CONFIG_DIR:-$HOME/.claude}" node -e '
const crypto = require("crypto")
console.log(crypto.createHash("sha256").update(process.env.CONFIG_DIR).digest("hex").slice(0, 8))
')

security find-generic-password -a "$(id -un)" \
  -s "Claude Code-credentials-$HASH" >/dev/null \
  && echo "hashed Keychain service exists"
```

With this state, OpenUsage reports:

```
Not logged in. Run `claude` to authenticate.
```

### Optional: reproduce from a machine that still has legacy credentials

If the legacy credential file or legacy Keychain service still exists, move them out of OpenUsage's lookup path first.

For the legacy file credential:

```
if [ -f ~/.claude/.credentials.json ]; then
  mv ~/.claude/.credentials.json ~/.claude/.credentials.json.bak
fi
```

For the legacy Keychain service, Keychain has no rename operation. The closest non-destructive flow is to copy the legacy secret to a backup service, then remove the original legacy service:

```
if security find-generic-password -a "$(id -un)" -s "Claude Code-credentials" >/dev/null 2>&1; then
  security add-generic-password -U \
    -a "$(id -un)" \
    -s "Claude Code-credentials.backup" \
    -w "$(security find-generic-password -w -a "$(id -un)" -s "Claude Code-credentials")"

  security delete-generic-password \
    -a "$(id -un)" \
    -s "Claude Code-credentials"
fi
```

### Expected Behavior

OpenUsage should detect the valid Claude Code login and show Claude usage.

### Actual Behavior

OpenUsage reports Session expired when a stale legacy Keychain item exists, and Not logged in after that stale legacy item is removed, even though Claude Code is logged in successfully.

### OpenUsage Version

0.6.21

### Operating System

macOS (Apple Silicon)

### Screenshots

Image

### Additional Context

If this turns out to be the right root cause, I am happy to open a PR shortly. The fix should be small: try the hashed `Claude Code-credentials-<hash>` service in addition to the existing legacy `Claude Code-credentials` lookup.

robinebers:
## Human Summary 👀

Issue #423 said that whenever `CLAUDE_CONFIG_DIR` is set, the keychain name changes. This PR fixes that by replicating exactly what Claude Code does.

## Summary

Fixes #423 — when `CLAUDE_CONFIG_DIR` is set, recent Claude Code stores macOS Keychain credentials under a hashed service name. The plugin only checked the unhashed name, so it reported `Not logged in` even on a freshly-authenticated machine.

## What changed
- `plugins/claude/plugin.js`: build a list of keychain service candidates and probe them in order — hashed first (when applicable), legacy second. Decompiled from the upstream `claude` binary, the service-name builder is:
  ```js
  function iN(H = "") {
    let _ = B6();  // (CLAUDE_CONFIG_DIR ?? path.join(homedir(), ".claude")).normalize("NFC")
    let K = !process.env.CLAUDE_CONFIG_DIR
      ? ""
      : `-${sha256(_).digest("hex").substring(0, 8)}`;
    return `Claude Code${OAUTH_FILE_SUFFIX}${H}${K}`;
  }
  ```
  Key insight: **the hash suffix is only appended when `CLAUDE_CONFIG_DIR` is explicitly set.** When unset, upstream uses the legacy unhashed `Claude Code-credentials`. The plugin now mirrors that exactly, including `.normalize("NFC")` on the hash input.

  The successful service name is also threaded through `loadCredentials` → `saveCredentials` so token refreshes write back to the same entry they were loaded from (no more accidentally writing to the legacy slot when the hashed one was the source).

- `src-tauri/src/plugin_engine/host_api.rs`: expose `host.crypto.sha256Hex(text)` to plugin JS (lowercase hex, matches Node's `crypto.createHash("sha256").update(x).digest("hex")`). New Rust dep: `sha2 = "0.10"`. **No env-allowlist changes** — `HOME` is not needed because we only hash when `CLAUDE_CONFIG_DIR` is set, and at that point we use its value directly.

## Why
The reporter (#423) traced this precisely: a re-auth created `Claude Code-credentials-<8-char-hash>` and the plugin's unhashed lookup found nothing. The hash matches `sha256(CLAUDE_CONFIG_DIR.normalize("NFC")).slice(0, 8)` — confirmed against the decompiled upstream `iN()` and `B6()` helpers.

## Test plan
- [x] Hashed entry found when `CLAUDE_CONFIG_DIR` is set (the #423 regression).
- [x] Legacy fallback works when `CLAUDE_CONFIG_DIR` is set but only the unhashed entry exists.
- [x] No hash computed when `CLAUDE_CONFIG_DIR` is unset (matches upstream behavior).
- [x] `USE_STAGING_OAUTH=1` produces `Claude Code-staging-oauth-credentials-<hash>` when paired with `CLAUDE_CONFIG_DIR`.
- [x] `CLAUDE_CONFIG_DIR=~/some-custom-claude-home` is hashed verbatim, no tilde expansion.
- [x] Rust unit test covers `sha256Hex` against a known vector.
- [ ] Manual smoke on the reporter's machine (CLAUDE_CONFIG_DIR set, only hashed keychain entry).

## Follow-ups
None planned — fix is contained.

### DIFF
--- a/plugins/claude/plugin.js
+++ b/plugins/claude/plugin.js
@@ -211,22 +211,22 @@
   }
 
   function computeKeychainHashSuffix(ctx) {
-    // Mirrors Claude Code's macOsKeychainHelpers.ts:
-    //   expandedConfigDir = process.env.CLAUDE_CONFIG_DIR (raw, NO tilde expansion)
-    //                       || ${HOME}/.claude
-    //   suffix            = sha256(expandedConfigDir).slice(0, 8)
+    // Mirrors upstream Claude Code (decompiled from the binary):
+    //   const suffix = !process.env.CLAUDE_CONFIG_DIR
+    //     ? ""
+    //     : "-" + sha256(CLAUDE_CONFIG_DIR.normalize("NFC")).slice(0, 8)
+    // The hash is ONLY appended when CLAUDE_CONFIG_DIR is explicitly set;
+    // when unset, upstream uses the legacy unhashed service name.
     const explicitConfigDir = readEnvText(ctx, "CLAUDE_CONFIG_DIR")
-    let expandedDir
-    if (explicitConfigDir) {
-      expandedDir = explicitConfigDir
-    } else {
-      const home = readEnvText(ctx, "HOME")
-      if (!home) return null  // can't compute hash; caller will fall back to legacy candidate only
-      expandedDir = home + "/.claude"
-    }
+    if (!explicitConfigDir) return null
     const sha256Hex = ctx.host && ctx.host.crypto && ctx.host.crypto.sha256Hex
     if (typeof sha256Hex !== "function") return null
-    const digest = sha256Hex(expandedDir)
+    // Match upstream's `.normalize("NFC")` exactly.
+    const normalized =
+      typeof explicitConfigDir.normalize === "function"
+        ? explicitConfigDir.normalize("NFC")
+        : explicitConfigDir
+    const digest = sha256Hex(normalized)
     if (typeof digest !== "string" || digest.length < 8) return null
     return digest.slice(0, 8)
   }
@@ -235,8 +235,8 @@
     const base = buildClaudeBaseKeychainService(ctx)
     const candidates = []
     const hash = computeKeychainHashSuffix(ctx)
-    if (hash) candidates.push(base + "-" + hash)  // newer Claude Code (hashed)
-    candidates.push(base)                          // legacy / compatibility
+    if (hash) candidates.push(base + "-" + hash)  // hashed (CLAUDE_CONFIG_DIR set)
+    candidates.push(base)                          // legacy / default
     return candidates
   }
 
--- a/plugins/claude/plugin.js
+++ b/plugins/claude/plugin.js
@@ -206,8 +206,38 @@
     }
   }
 
-  function getClaudeKeychainService(ctx) {
+  function buildClaudeBaseKeychainService(ctx) {
     return KEYCHAIN_SERVICE_PREFIX + getOauthConfig(ctx).oauthFileSuffix + "-credentials"
+  }
+
+  function computeKeychainHashSuffix(ctx) {
+    // Mirrors Claude Code's macOsKeychainHelpers.ts:
+    //   expandedConfigDir = process.env.CLAUDE_CONFIG_DIR (raw, NO tilde expansion)
+    //                       || ${HOME}/.claude
+    //   suffix            = sha256(expandedConfigDir).slice(0, 8)
+    const explicitConfigDir = readEnvText(ctx, "CLAUDE_CONFIG_DIR")
+    let expandedDir
+    if (explicitConfigDir) {
+      expandedDir = explicitConfigDir
+    } else {
+      const home = readEnvText(ctx, "HOME")
+      if (!home) return null  // can't compute hash; caller will fall back to legacy candidate only
+      expandedDir = home + "/.claude"
+    }
+    const sha256Hex = ctx.host && ctx.host.crypto && ctx.host.crypto.sha256Hex
+    if (typeof sha256Hex !== "function") return null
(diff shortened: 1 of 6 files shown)


robinebers:
Course-correction in 5651a3a: the hash suffix is **only** appended when `CLAUDE_CONFIG_DIR` is explicitly set — confirmed by decompiling the upstream `claude` binary. My first commit also hashed the default `${HOME}/.claude` case, which upstream doesn't do; that meant we'd look up a keychain entry that never exists. Removed the `HOME` dependency and the corresponding env-allowlist entry, added `.normalize("NFC")` to match upstream exactly, and updated tests.

V1CeVersaa:
This works perfectly on my machine. Thank you for your contribution!

validatedev:
@codex review

validatedev (approved):

PR #427, opened 2026-05-01
feat: add 12h/24h/auto time format setting

### LINKED ISSUE #426: Setting to override clock format (12h / 24h)
### Problem

Reset times in OpenUsage (e.g. `Resets today at 11:04`) are formatted via `Intl.DateTimeFormat` with no `hour12` flag, so the clock format is dictated entirely by the user's OS locale. There's no way to override it.

This is a problem for:
- Users in 12-hour locales who prefer 24-hour clocks (or vice versa).
- Users in mixed environments (e.g. UK locale with US date conventions, or developers who want consistency with their terminal/IDE).
- Anyone whose system locale doesn't match how they want OpenUsage to display time.

Today the only workaround is to change the OS locale itself, which affects every other app.

### Proposed Solution

Add a **Time Format** section to Settings (next to **Reset Timers**) with three options:

- **Auto** (default) — follows OS locale, matches current behavior.
- **12-hour** — forces `11:04 AM` style.
- **24-hour** — forces `11:04` style.

Each option's button shows a live preview of how times will render. Persisted via the existing Tauri store.

### Alternatives Considered

- **Auto-detect only, no override** — current behavior. Doesn't address users who want a different format than their OS.
- **Single 12h/24h toggle** without `Auto` — forces every user to make an explicit choice and can't track locale changes.
- **Inferring from a separate locale setting** — heavier; OpenUsage doesn't have a locale concept beyond what `Intl` provides.

The 3-option `Auto / 12h / 24h` pattern is the standard across iOS, macOS, GNOME, etc., and keeps the default a no-op.

HDash:
## Description

Adds a Time Format setting (Auto / 12-hour / 24-hour) to Settings, below Reset Timers. Default **Auto** preserves existing locale-driven behavior; explicit choices override.

The mode flows from the preferences store through the reset-tooltip formatter, so reset times in provider cards and tooltips render in the user's preferred clock.

## Related Issue

Closes #426

## Type of Change

- [x] New feature

## Testing

- [x] `npm run build` (tsc) — clean
- [x] `npm test` — 60 files, 1087 tests passing
- [x] Manually tested in `tauri dev` (see screenshots)

## Screenshots

**Before** — current `main`, no time format option:

before

**After** — new Time Format section in Settings:

settings

**After** — provider card rendering reset time in the chosen format:

main

## Checklist

- [x] No new dependencies introduced


HDash:
Thanks @copilot — addressed in 456d8d2:

- **`TIME_FORMAT_KEY` naming** — renamed the stored key value from `"timeFormat"` to `"timeFormatMode"` to match `themeMode` / `displayMode` / `resetTimerDisplayMode`. Test storage stubs updated. No migration needed since the key has not shipped.
- **Locale-dependent assertions in `reset-tooltip.test.ts`** — rewrote the new tests to derive expected strings via `getTimeFormatter("12h" | "24h").format(...)` and assert equality, instead of matching literal `"AM/PM"` / `"14:05"`. Two of the original three tests were consolidated into one since they now share the same setup; the differ-sanity check (`expected12 !== expected24`) preserves the intent of verifying the formats actually diverge.
- **Missing `setMenubarIconStyle` stub in `useSettingsDisplayActions` tests (3 sites)** — added `setMenubarIconStyle: vi.fn()` to the main, toggle, and error-logging test invocations. Pre-existing gap (test files are excluded from `tsc`, so this was silent), but agreed it's safer to keep the test inputs representative of the hook's full signature.

Local: `tsc --noEmit` clean, `vitest run` 60 files / 1086 tests passing.

validatedev:
@codex review

validatedev (approved):
No issues. Looks good to me, and feels like a good improvement. Thanks!

validatedev:
@codex review again

HDash:
Hey @validatedev, appreciate the quick reviews!

That said, there's a growing pile of PRs that have been reviewed by you (so one reviewer) but never merged. Given this is a side project and everyone's stretched, would it make sense to loosen the two-reviews rule a bit? Feels like it might be the bottleneck here.

validatedev:
@HDash welcome! this is not in my control, sorry about that. @robinebers will also check whenever he's available. we always want the best possible outcome for the project, thanks for the understanding

validatedev (approved):

robinebers (approved):
LGTM

Issue #488, opened 2026-05-22
Pause useNowTicker while the panel WebView is hidden


zergzorg:
## Problem

`useNowTicker` drives UI-only time labels such as relative timestamps and short cooldown/countdown text. Today those timers can keep React waking up even when the Tauri panel WebView is hidden.

OpenUsage intentionally keeps background refreshes working on macOS, so this is not about re-enabling App Nap or stopping provider probes. The narrow target is to avoid running display-only React intervals while there is no visible panel to update.

This is a small, separable frontend optimization that complements #487.

## Proposed change

Add visibility-aware pausing to `useNowTicker`:

- default to pausing the interval when `document.hidden === true`
- resume and refresh `now` on `visibilitychange` when the panel becomes visible again
- keep an escape hatch such as `pauseWhenHidden: false` for any caller that truly needs ticking while hidden
- keep the existing `enabled`, `intervalMs`, `stopAfterMs`, and `resetKey` behavior intact

Candidate scope:

- `src/hooks/use-now-ticker.ts`
- `src/hooks/use-now-ticker.test.ts`

## Why this helps

The app currently has UI timers that are useful while the menu panel is open, but do not add value while the panel is hidden. Pausing them reduces idle wakeups and React state updates without changing provider refresh frequency or user-visible data freshness.

## Acceptance criteria

- `useNowTicker` does not start/continue its interval while `document.hidden` is true by default.
- The hook updates immediately after `visibilitychange` makes the document visible again.
- Existing disabled and `stopAfterMs` behavior remains covered by tests.
- A caller can opt out with `pauseWhenHidden: false`.
- No changes to provider probe scheduling or background refresh behavior.

## Test plan

- Add/extend hook tests covering hidden document pause and visible resume.
- Run the focused test file, for example:

```bash
vitest run src/hooks/use-now-ticker.test.ts
```

- Run the regular app build/test command used by the project before opening a PR.

zergzorg:
Implemented by #490 and shipped in `v0.6.25`.

That PR pauses `useNowTicker` while the panel document is hidden, keeps the visible-panel tick behavior unchanged, and refreshes immediately when the panel becomes visible again. This issue can be closed unless there is a remaining edge case to track separately.

robinebers:
The complete rewrite of OpenUsage (v0.7.0) is coming shortly. With that, we're cleaning up the repo from old issues that no longer apply. For more information, check https://github.com/robinebers/openusage/issues/565

Issue #487, opened 2026-05-22
Resource budget for background refreshes: adaptive scheduler, probe limits, and diagnostics


zergzorg:
### Problem

This is not a request for shorter polling intervals, and not a duplicate of the existing point fixes around memory leaks, orphaned `ccusage` processes, or provider 429s.

OpenUsage is a menu bar app that is meant to run for days. That makes the background resource budget itself worth tracking: CPU wakeups, hidden WebView work, overlapping probes, external child processes, local HTTP API threads, and slow provider diagnostics.

Related but narrower issues/PRs:

- #400: current open report for severe memory growth while running in the background.
- #399 and #375: closed high-memory reports, apparently hard to reproduce and possibly related to slow/stuck provider probes.
- #431 / #433: fixed orphaned `ccusage` process accumulation with process group kill/timeout handling.
- #377 / #378: fixed Claude 429 handling with `Retry-After` support and cached usage.
- #94: stale but still relevant idea to stop polling exhausted providers until reset time.
- #192 / #382 / #134: requests for more frequent polling were rejected/closed because OpenUsage should not hammer provider APIs.

This issue is about an app-wide resource budget and observability layer so OpenUsage can stay lightweight during long background sessions while still refreshing reliably.

### Current observations

On my machine with OpenUsage 0.6.24, idle CPU is low in a short sample, so this is not a claim that the current release is always leaking:

```text
OpenUsage main process sample:
- CPU: 0.0-0.2% during a short idle `top` sample
- Physical footprint from `sample`: 24.2 MB, peak 28.0 MB
- WebKit child processes are expected for Tauri/WebKit
```

But there are several code paths where background work is intentionally kept alive or currently unbounded:

- `src-tauri/src/app_nap.rs` disables App Nap for the app lifetime.
- `src-tauri/src/webkit_config.rs` sets WebKit `inactiveSchedulingPolicy` to `None`, so hidden WebView timers can continue running.
- `src/hooks/app/use-probe-auto-update.ts` schedules auto refreshes in the frontend WebView.
- `src/hooks/use-now-ticker.ts`, `src/components/panel-footer.tsx`, and `src/components/provider-card.tsx` can generate timer-driven React work for countdown/reset/cooldown display.
- `src/hooks/app/use-panel.ts` uses ResizeObserver/MutationObserver and calls window resize logic when plugin-derived view state changes.
- `src/hooks/app/use-tray-icon.ts` can rasterize tray icons via canvas after probe results.
[...]

PR #490, opened 2026-05-22
[codex] Pause ticker while panel is hidden

zergzorg:
## Summary

- pause `useNowTicker` intervals while `document.hidden` is true by default
- refresh `now` immediately when the document becomes visible again, before the next interval tick
- add hook coverage for initially hidden documents, active ticker pause/resume, and opt-out behavior

## Why

`useNowTicker` drives display-only UI such as countdowns and relative timestamps. When the menu panel WebView is hidden, those React timer updates do not change visible UI but can still wake the app. Pausing the ticker while hidden reduces idle UI work without changing provider probes or background refresh scheduling.

This addresses #488 and complements the broader resource-budget discussion in #487.

## Notes

The default behavior is visibility-aware, but callers can keep the old hidden ticking behavior with `pauseWhenHidden: false` if a future use case needs it.

## Validation

- `node ./node_modules/vitest/vitest.mjs run src/hooks/use-now-ticker.test.ts`
- `bun run build`

The full Vitest run passes. Existing tests print Tauri store mock stderr in `App.test.tsx`, but the suite exits successfully.

---
## Summary by cubic
Pause `useNowTicker` ticks while the document is hidden to reduce idle updates, and resume with an immediate refresh when it becomes visible. Adds a `pauseWhenHidden` option (default true), respects `enabled` on visibility changes (no refresh when disabled), and tests for initial hidden state, pause/resume, disabled behavior, and opt‑out.

<sup>Written for commit be7d219de3a68e3e1369c4bba2ad4c06552e08ad. Summary will update on new commits. <a href="https://cubic.dev/pr/robinebers/openusage/pull/490?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/hooks/use-now-ticker.test.ts
+++ b/src/hooks/use-now-ticker.test.ts
@@ -1,13 +1,20 @@
 import { renderHook, act } from "@testing-library/react"
-import { afterEach, describe, expect, it, vi } from "vitest"
+import { afterEach, beforeAll, describe, expect, it, vi } from "vitest"
 import { useNowTicker } from "./use-now-ticker"
 
 describe("useNowTicker", () => {
+  let originalDocumentHiddenDescriptor: PropertyDescriptor | undefined
+
+  beforeAll(() => {
+    originalDocumentHiddenDescriptor = Object.getOwnPropertyDescriptor(document, "hidden")
+  })
+
   afterEach(() => {
-    Object.defineProperty(document, "hidden", {
-      configurable: true,
-      value: false,
-    })
+    if (originalDocumentHiddenDescriptor) {
+      Object.defineProperty(document, "hidden", originalDocumentHiddenDescriptor)
+    } else {
+      Reflect.deleteProperty(document, "hidden")
+    }
     vi.useRealTimers()
   })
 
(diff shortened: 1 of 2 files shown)


zergzorg:

This PR is independent from the #487 probe-resource stack. It belongs to #488 and only reduces UI-only ticker work while the panel WebView is hidden.

Suggested merge position: merge anytime. It can land before or after #498/#499/#500 because it does not change probe scheduling, backend probe execution, or provider runtime behavior.

validatedev:
@codex review

zergzorg:

- gated the visibility-resume `setNow(Date.now())` behind `enabled`, so disabled tickers stay stable;
- restored the original `document.hidden` descriptor after tests;
- added a regression test for disabled + visibility resume.

Verified locally with `node ./node_modules/vitest/vitest.mjs run src/hooks/use-now-ticker.test.ts` using the bundled Node runtime. The updated branch is also rebased on current `main`, and CI is green.

robinebers (approved):
LGTM. Display-only ticker change — no impact on app self-update checks (`useAppUpdate` own setInterval), plugin auto-refresh (`useProbeAutoUpdate` own setInterval), or tray icon updates (`useTrayIcon` is probe-result-driven). macOS App Nap is already disabled in `app_nap.rs` so background timers keep firing while the panel is hidden. Tests are solid, both bot-review nits addressed in commit 2.

PR #497, opened 2026-05-23
fix(cursor): add REST fallback for Free accounts and fix team inference

deep5deep:
This pull request addresses an issue where Cursor Free tier users fail the usage probe in OpenUsage.

### Problems Addressed:
1. **Missing limit in Connect API:** For Cursor Free tier accounts, the internal Connect RPC endpoint (`GetCurrentPeriodUsage`) does not return a `limit` property in `planUsage`.
2. **False Positive Team Account:** The plugin's `isTeamAccount` check matched Free tier users because `spendLimitUsage.pooledLimit` is `0` (which is a number). This caused the plugin to incorrectly attempt a Team/REST fallback.
3. **No REST Usage Limits:** The REST request-based fallback threw a `request-based: no usage data available` exception because Free tier users have no fast requests limits in the REST API.

### Changes:
- **Added REST & Stripe Fallbacks:** Added a fallback mechanism inside the `probe` function so that if the main Connect API request fails (e.g., due to TLS/HTTP2 handshake issues on some systems), it gracefully resolves the Free plan name and total requests using the REST and Stripe APIs.
- **Fixed Team Inference check:** Changed the check so that `pooledLimit` must be greater than `0` to prevent matching Free accounts:
  ```javascript
  (su && typeof su.pooledLimit === 'number' && su.pooledLimit > 0)
  ```

---
## Summary by cubic
Adds REST + Stripe fallback in the Cursor usage probe so Free accounts show plan and usage when the Connect API fails or returns no limits. Fixes team inference for Free users and improves error handling; also extracts a shared `fetchStripe` helper.

- **New Features**
  - Fallback resolves plan from Stripe membership and usage from REST; shows progress if a limit exists, otherwise shows used count or “Free tier active”.
  - Adds student verification label when available.

- **Bug Fixes**
  - Team detection now requires pooledLimit > 0, preventing Free accounts from matching Team.
  - Downgrades Connect request errors to warnings with clearer messages (including “after refresh”), preserving fallback instead of failing the probe.

<sup>Written for commit 4f617d5df95e05967a720038d29b5b212ec8548b. Summary will update on new commits. <a href="https://cubic.dev/pr/robinebers/openusage/pull/497?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/plugins/cursor/plugin.js
+++ b/plugins/cursor/plugin.js
@@ -275,7 +275,7 @@
     }
   }
 
-  function fetchStripeBalance(ctx, accessToken) {
+  function fetchStripe(ctx, accessToken) {
     var session = buildSessionToken(ctx, accessToken)
     if (!session) {
       ctx.host.log.warn("stripe: cannot build session token")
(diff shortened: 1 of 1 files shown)


validatedev:
@codex review

zergzorg:
I took a local pass over the unresolved review feedback and reproduced the issue.

Current PR head fails the Cursor plugin test file:

- `bun run test plugins/cursor/plugin.test.js --run`
- result before fix: 5 failures, all caused by the new `connectApiFailed` catch swallowing auth/request errors and returning the REST/Stripe fallback instead of throwing.

The smallest fix I found is to remove the synthetic fallback for exceptions from `retryOnceOnAuth` and preserve the existing error contract. The existing request-based fallback later in the probe still handles missing/limitless `planUsage`; this just prevents auth/network/request exceptions from being misreported as a successful Free result.

Patch that passes locally:

```diff
diff --git a/plugins/cursor/plugin.js b/plugins/cursor/plugin.js
@@
-    let connectApiFailed = false
     let didRefresh = false
@@
           } catch (e) {
             ctx.host.log.warn("connectPost failed: " + String(e))
-            throw e
+            if (didRefresh) {
+              throw "Usage request failed after refresh. Try again."
+            }
+            throw "Usage request failed. Check your connection."
           }
@@
     } catch (e) {
-      ctx.host.log.warn("Connect API failed, entering REST fallback: " + String(e))
-      connectApiFailed = true
-    }
-
-    if (connectApiFailed) {
-      ...
-      return { plan: plan, lines: lines }
+      if (typeof e === "string") throw e
+      ctx.host.log.error("usage request failed: " + String(e))
+      throw "Usage request failed. Check your connection."
     }
```

After applying that locally, `plugins/cursor/plugin.test.js` passes: 59/59.

This should close the Codex P1 and the related Copilot comments: auth/session string errors are re-thrown, non-string request wrapper errors stay user-facing failures, and the PR no longer defaults to `Free` on a failed Stripe/REST fallback after a Connect exception.

deep5deep:
Fixed per Copilot review feedback.

**Root cause:** The `connectApiFailed` catch block was swallowing *all* thrown values — including string auth errors like `"Token expired. Sign in via…"` — and falling back to a REST/Stripe path that returned misleading Free-tier data instead of prompting re-authentication.

**Fix applied:**
- String errors (auth messages thrown from `refreshToken`) are now re-thrown immediately via `if (typeof e === "string") throw e`
- Non-string transport/network errors are wrapped as `"Usage request failed [after refresh]. Check your connection."`  
- The entire `connectApiFailed` REST fallback block has been removed — the existing `planUsageLimitMissing` path already handles Free tier accounts correctly when the Connect API returns a 200 response without a `limit`

All 59 tests pass locally.

PR #498, opened 2026-05-23
[codex] Skip auto-update probes already in flight

zergzorg:
## Summary

- skip auto-update probes for providers that are already marked `loading`
- keep refreshing idle providers when only part of the enabled set is still in flight
- keep the next auto-update schedule moving even when every provider is still busy
- add focused hook coverage and update App integration tests to model initial probe completion before auto-update

## Why

This is a small first slice of #487. If a provider probe is slow or stuck, the fixed auto-update interval should not start another probe for that same provider on top of the existing one. The app already uses `loading` to prevent repeated manual refreshes; this applies the same resource-budget guard to background auto-updates.

The change does not alter the configured refresh interval, provider plugin runtime, or manual refresh behavior. It only avoids redundant auto-update work for providers that are still in flight.

## Validation

- `node ./node_modules/vitest/vitest.mjs run src/hooks/app/use-probe-auto-update.test.ts`
- `node ./node_modules/vitest/vitest.mjs run src/App.test.tsx -t "auto-update"`
- `bun run build`

The full Vitest run passes. Existing App tests still print Tauri store mock stderr, but the suite exits successfully.

Closes part of #487.

---
## Summary by cubic
Skip auto-update probes already in flight and keep probe state in sync so loading checks are accurate. Only idle providers run; the schedule still advances when all providers are busy. Part of #487.

- **Bug Fixes**
  - Only start auto-update batches for idle providers; skip those still `loading`.
  - If all enabled providers are busy, do not start a batch and reschedule the next run.
  - Keep `pluginStatesRef` current when setting loading/errors so `isPluginLoading` reflects changes immediately; wire `isPluginLoading` into `use-probe-auto-update` and update tests (App timers, initial batch completion, and new `use-probe-state` coverage).

<sup>Written for commit 6e414966d0aae0c54beada14f2a3d1b76c8ff1c8. Summary will update on new commits. <a href="https://cubic.dev/pr/robinebers/openusage/pull/498?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src/App.test.tsx
+++ b/src/App.test.tsx
@@ -343,6 +343,7 @@ describe("App", () => {
   })
 
   afterEach(() => {
+    vi.useRealTimers()
     delete (HTMLElement.prototype as unknown as Record<string, unknown>).scrollHeight
   })
 
@@ -1373,6 +1374,13 @@ describe("App", () => {
     await vi.waitFor(() => expect(state.startBatchMock).toHaveBeenCalled())
 
     // Clear the initial batch call count
+    state.probeHandlers?.onResult({
+      providerId: "a",
+      displayName: "Alpha",
+      iconUrl: "icon-a",
+      lines: [{ type: "text", label: "Now", value: "OK" }],
+    })
+    state.probeHandlers?.onBatchComplete()
     const initialCalls = state.startBatchMock.mock.calls.length
 
     // Advance time by 5 minutes to trigger the interval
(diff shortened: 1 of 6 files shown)


zergzorg:

Recommended #487 merge order:

1. #498 first: skip auto-update probes for providers already in flight.

This PR has no hard dependency on #499 or #500 and is the smallest behavioral guard in the stack. It is the best first merge because it prevents redundant background work before the backend-level limits are reviewed.

validatedev:
@codex review

zergzorg:

- `useProbeState` now updates `pluginStatesRef.current` synchronously before publishing React state;
- added a regression test that confirms `setLoadingForPlugins()` is immediately visible through the ref, which closes the same-tick race for `isPluginLoading`.

Verified locally with `node ./node_modules/vitest/vitest.mjs run src/hooks/app/use-probe-state.test.ts src/hooks/app/use-probe-auto-update.test.ts` using the bundled Node runtime. The updated branch is also rebased on current `main`, and CI is green.

robinebers (approved):
LGTM. Clean focused change. The synchronous `pluginStatesRef` update via `updatePluginStates` is the right way to make `isPluginLoading` actually correct on the next interval tick. Tests cover the partial-idle and all-loading branches. Verified all 78 App tests + new hook tests pass locally.

PR #499, opened 2026-05-23
[codex] Cap concurrent plugin probes per batch

zergzorg:
## Summary

- cap each probe batch to at most 4 concurrent plugin probe workers
- keep queued plugins inside the same batch instead of spawning every enabled plugin at once
- preserve the existing `probe:result` and `probe:batch-complete` event flow
- add a small unit test for the worker-count cap

## Why

This is another small slice of #487. A batch can include many enabled providers, and the previous implementation spawned one blocking task per plugin immediately. That is simple, but it can create avoidable CPU/thread/process pressure when several plugins do filesystem, network, SQLite, keychain, or helper-process work at the same time.

The new queue keeps provider results asynchronous but limits batch fan-out, so background refreshes have a bounded resource profile.

## Notes

The cap is intentionally conservative and local to backend batch execution. It does not change polling intervals, provider enablement, manual refresh behavior, or the frontend event contract.

## Validation

- `bun run bundle:plugins`
- `cargo test probe_worker_count_is_bounded`
- `bun run build`

I also ran full `cargo test`; Rust compilation and 96 tests passed, but an existing ccusage process-timeout test fails locally while trying to read `descendant.pid` before it exists. I did not change that unrelated test in this PR.

Part of #487.

---
## Summary by cubic
Cap concurrent plugin probes per batch to 4 to bound CPU and thread usage. Results stay async and `probe:result`/`probe:batch-complete` events are unchanged.

- **Refactors**
  - Use a small worker pool per batch and queue remaining plugins inside the batch.
  - Cache successful outputs as before; no changes to polling intervals, provider enablement, manual refresh, or the frontend contract.
  - Add `probe_worker_count_is_bounded` test to verify the cap.

<sup>Written for commit 7d3442f79795fadead86ade427312ebb8f1ee754. Summary will update on new commits. <a href="https://cubic.dev/pr/robinebers/openusage/pull/499?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/src-tauri/src/lib.rs
+++ b/src-tauri/src/lib.rs
@@ -8,7 +8,7 @@ mod tray;
 #[cfg(target_os = "macos")]
 mod webkit_config;
 
-use std::collections::{HashMap, HashSet};
+use std::collections::{HashMap, HashSet, VecDeque};
 use std::path::PathBuf;
 use std::sync::atomic::{AtomicUsize, Ordering};
 use std::sync::{Arc, Mutex, OnceLock};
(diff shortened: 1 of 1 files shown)


zergzorg:

Recommended #487 merge order:

1. #498 first: skip duplicate auto-update probes for providers already in flight.

This PR is based on `main` and has no hard code dependency on #498 or #500. It is recommended after #498 conceptually: once duplicate provider work is avoided, this caps how many different providers a batch can run at once.

validatedev:
@codex review

robinebers (approved):
LGTM

PR #500, opened 2026-05-23
[codex] Add per-probe runtime deadline

zergzorg:
## Summary

- add a 30s wall-clock budget for each plugin probe execution
- use the QuickJS interrupt handler to stop CPU-bound plugin scripts after the probe budget expires
- pass the same deadline into the host API for the currently deadline-aware blocking calls: `host.http` and `host.ccusage`
- keep the existing plugin output contract by returning an Error badge when a probe times out

## Why

This is another small slice of #487. Previous follow-ups limit duplicate auto-update probes and batch fan-out; this one bounds the duration of an individual provider probe for the highest-risk paths: CPU-bound JavaScript loops and the existing long-running HTTP/ccusage calls.

The timeout is intentionally conservative at 30 seconds per provider. Existing provider-level HTTP defaults and ccusage timeouts continue to work, but they cannot exceed the remaining per-probe budget.

## Scope and limitations

This is not a complete hard deadline for every possible host API path yet. It does not rewrite every synchronous local operation into a killable subprocess helper. In particular, local filesystem operations and the remaining system-command based helpers (`security`, `sqlite3`, `ps`, `lsof`, plus env shell reads) are left for a later, more invasive slice.

That follow-up is called out separately so this PR stays small and reviewable: first stop CPU-bound plugin scripts and clamp the known long-running HTTP/ccusage calls, then add a shared timeout wrapper for the remaining host subprocess helpers.

This does not change frontend scheduling, event names, provider enablement, or manual refresh behavior.

## Validation

- `bun run bundle:plugins`
- `cargo test plugin_engine::runtime::tests`
- `cargo test plugin_engine::host_api::tests::probe_deadline_clamps_host_timeout_to_remaining_budget`
- `cargo test plugin_engine::host_api::tests::ccusage_timeout_stops_runner_fallback`
- `cargo test plugin_engine::host_api::tests::ccusage_runner_retries_legacy_package_when_current_package_fails`
- `bun run build`

I also ran full `cargo test`; Rust compilation and 97 tests passed, but the existing `ccusage_timeout_kills_descendant_and_closes_pipes` test still fails locally while trying to read `descendant.pid` before it exists. I did not change that unrelated test in this PR. That test is handled separately in #501.

Part of #487.

### DIFF
--- a/src-tauri/src/plugin_engine/host_api.rs
+++ b/src-tauri/src/plugin_engine/host_api.rs
@@ -31,6 +31,7 @@ const WHITELISTED_ENV_VARS: [&str; 16] = [
     "SYNTHETIC_API_KEY",
     "PI_CODING_AGENT_DIR",
 ];
+const MIN_BLOCKING_TIMEOUT: Duration = Duration::from_millis(1);
 
 #[derive(Clone, Copy, Debug)]
 pub(crate) struct ProbeDeadline {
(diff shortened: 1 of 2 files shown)


zergzorg:

Recommended #487 merge order:

1. #498 first: skip duplicate auto-update probes for providers already in flight.

This PR is based on `main` and has no hard code dependency on #498 or #499, but it is easiest to review last: after duplicate work and batch width are bounded, this bounds the duration of a single provider probe.

Remaining follow-up candidates after this stack: shared timeout wrapper for remaining host subprocess calls, bounded local HTTP API accept path, cache-write debounce, and lightweight probe diagnostics.

validatedev:
@codex review

zergzorg:

- `ProbeDeadline::clamp_duration` now returns no timeout when the remaining budget is already exhausted instead of inventing an extra 1ms;
- `host.http` and `host.ccusage` now short-circuit as timed out in that case before starting blocking work;
- added coverage for the elapsed-budget case.

Verified locally with `cargo test probe_deadline`, `cargo test run_probe_times_out_cpu_bound_script`, and `cargo test ccusage_timeout_stops_runner_fallback`. The updated branch is also rebased on current `main`, and CI is green.

robinebers (approved):
LGTM

PR #502, opened 2026-05-23
[codex] Bound local HTTP API concurrency

zergzorg:
## Summary

- cap the local HTTP API to 16 active connection workers
- return `503 {"error":"server_busy"}` without spawning another worker when the cap is reached
- add write timeouts alongside the existing read timeout
- add focused tests for permit release and the busy response

## Why

This is a small follow-up slice of #487. The local API currently accepts each TCP connection and immediately spawns an OS thread for it. Under accidental or intentional local stress traffic, that can grow worker threads without a local app-side bound.

The new connection limiter keeps normal local API behavior unchanged for regular clients, while making overload explicit and bounded. This is intentionally independent from the probe-resource PR stack (#498/#499/#500) and avoids touching `host_api.rs`, which already has active PRs.

## Notes

The local API is read-only and localhost-bound, so returning `503 server_busy` during overload is preferable to creating unbounded OS threads. The cap is deliberately conservative; normal use should be one or a few concurrent requests.

## Validation

- `cargo test local_http_api::server::tests`
- `bun run build`

I skipped `ccusage_timeout_kills_descendant_and_closes_pipes` in the broad Rust run because that unrelated flaky test is fixed separately in #501 but is not in `main` yet.

Part of #487.

---
## Summary by cubic
Bound the local HTTP API to 16 concurrent connections and return 503 {"error":"server_busy"} without spawning a worker when at capacity. Added 5s read/write timeouts, tests for permit release and 503 JSON/headers, and updated docs to include `server_busy` with backoff guidance.

<sup>Written for commit 0708889e4b2065254a65be70c380ca1250f5809f. Summary will update on new commits. <a href="https://cubic.dev/pr/robinebers/openusage/pull/502?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/docs/local-http-api.md
+++ b/docs/local-http-api.md
@@ -92,4 +92,6 @@ Error responses use this shape:
 }
 ```
 
-Possible error codes: `provider_not_found`, `not_found`, `method_not_allowed`.
+Possible error codes: `provider_not_found`, `not_found`, `method_not_allowed`, `server_busy`.
+
+`server_busy` returns **503 Service Unavailable** when the local API is already handling the maximum number of concurrent connections. Clients should back off and retry later.
--- a/src-tauri/src/local_http_api/server.rs
+++ b/src-tauri/src/local_http_api/server.rs
@@ -9,7 +9,6 @@ const BIND_ADDR: &str = "127.0.0.1:6736";
 const MAX_CONCURRENT_CONNECTIONS: usize = 16;
 const CONNECTION_TIMEOUT: Duration = Duration::from_secs(5);
 
-#[derive(Clone)]
 struct ConnectionLimiter {
     active: Arc<AtomicUsize>,
     max: usize,
--- a/src-tauri/src/local_http_api/server.rs
+++ b/src-tauri/src/local_http_api/server.rs
@@ -1,8 +1,61 @@
 use super::cache::{cache_state, enabled_snapshots_ordered};
 use std::io::{Read, Write};
 use std::net::{TcpListener, TcpStream};
+use std::sync::atomic::{AtomicUsize, Ordering};
+use std::sync::Arc;
+use std::time::Duration;
 
 const BIND_ADDR: &str = "127.0.0.1:6736";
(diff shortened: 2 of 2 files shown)


validatedev:
@codex review

robinebers:
Another great contribution. Thank you so much.

robinebers (changes requested):
Hey! 👋 This is Rob's AI reviewer. Thanks for the careful follow-up to #487!

Really clean change overall — atomic CAS loop is correct, Drop-based permit release means no leaks, the 503 path preserves CORS/`Connection: close`, and the new tests cover both the limiter invariant and the response shape. Two small things before merge:

1. **Document the new 503 in `docs/local-http-api.md`** — the "Error Responses" section currently lists `provider_not_found`, `not_found`, `method_not_allowed` as the possible error codes. Worth adding `server_busy` and a short note that the server may return `503 Service Unavailable` when the concurrent-connection cap is reached, so API consumers know to back off and retry rather than treat it as a hard failure.

2. **Drop the unused `#[derive(Clone)]` on `ConnectionLimiter`** — the limiter is constructed once in `start_server` and never cloned (permits hold their own `Arc<AtomicUsize>`), so the derive is dead. Tiny nit.

Everything else (timeouts, conservative cap, scope staying out of `host_api.rs`, test coverage) looks great.

zergzorg:
Addressed both requested changes in the latest push:

- documented server_busy in docs/local-http-api.md as a 503 Service Unavailable response that callers should back off and retry;
- removed the unused Clone derive from ConnectionLimiter.

Verified locally with:

- rustfmt --check src/local_http_api/server.rs
- cargo test local_http_api::server::tests
- bun run build
- git diff --check

Note: bun run build succeeds, but Vite prints the existing Node 22.0.0 version warning.

zergzorg:
Follow-up: updated the branch with current origin/main using a normal merge commit, no force-push.

Re-verified after the merge:

- cargo test local_http_api::server::tests
- bun run build

The Vite Node 22.0.0 warning is still present, but the build exits successfully.

PR #508, opened 2026-05-24
fix(antigravity): fix support for antigravity ide (v2)

kvanzuijlen:
## Description

This PR adds support for Antigravity IDE (the new name for what was previously known as Antigravity)

For backwards compatibility, it checks both the v2 (Antigravity IDE) and v1 (Antigravity) paths.

### Note on Plugin Renaming & Standalone v2
Google made some changes to Antigravity. Antigravity now refers to a standalone, agent-first app, while Antigravity IDE is now what was previously known as Antigravity IDE. I think it might make sense to rename the plugin to Antigravity IDE. I decided to keep it out of scope for this PR, as I suppose renaming the plugin could break existing users' configuration.

The standalone app is no longer VSC based, and thus doesn't have a state.vscdb.

## Related Issue

None.

## Type of Change

- [x] Bug fix
- [ ] New feature
- [ ] New provider plugin
- [ ] Documentation
- [ ] Performance improvement
- [ ] Other (describe below)

## Testing

- [x] I ran `bun run test` and all tests pass (verified with `npx vitest run`)

## Checklist

- [x] I read CONTRIBUTING.md

---
## Summary by cubic
Restore support for the renamed Antigravity IDE (v2) by reading OAuth tokens from the new v2 DB first and detecting the renamed IDE process. Docs now show the v2 path and the fallback order.

- **Bug Fixes**
  - Read tokens from `~/Library/Application Support/Antigravity IDE/User/globalStorage/state.vscdb` (v2), then fall back to the v1 path.
  - Skip missing DBs and fall back when v2 is corrupt or has no token.
  - Update LS discovery to look for both "antigravity" and "antigravity-ide" markers.

<sup>Written for commit e417c511ab12364248018ba4db616a8968d23ec8. Summary will update on new commits. <a href="https://cubic.dev/pr/robinebers/openusage/pull/508?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/docs/providers/antigravity.md
+++ b/docs/providers/antigravity.md
@@ -158,7 +158,9 @@ Interestingly, non-Google models (Claude, GPT-OSS) are proxied through Codeium/W
 
 Antigravity stores auth credentials in a VS Code-compatible state database.
 
-- **Path:** `~/Library/Application Support/Antigravity/User/globalStorage/state.vscdb`
+- **Paths (tried in order):**
+  - `~/Library/Application Support/Antigravity IDE/User/globalStorage/state.vscdb` (v2 - renamed app)
+  - `~/Library/Application Support/Antigravity/User/globalStorage/state.vscdb` (v1 - pre-rename, retained for backwards compatibility)
 - **Table:** `ItemTable` (`key` TEXT, `value` TEXT)
 
 ### antigravityUnifiedStateSync.oauthToken (sentinel envelope → protobuf)
--- a/plugins/antigravity/plugin.js
+++ b/plugins/antigravity/plugin.js
@@ -1,6 +1,7 @@
 (function () {
   var LS_SERVICE = "exa.language_server_pb.LanguageServerService"
-  var STATE_DB = "~/Library/Application Support/Antigravity/User/globalStorage/state.vscdb"
+  var STATE_DB_V1 = "~/Library/Application Support/Antigravity/User/globalStorage/state.vscdb"
+  var STATE_DB_V2 = "~/Library/Application Support/Antigravity IDE/User/globalStorage/state.vscdb"
   var CLOUD_CODE_URLS = [
     "https://daily-cloudcode-pa.googleapis.com",
     "https://cloudcode-pa.googleapis.com",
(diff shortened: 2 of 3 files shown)


robinebers (changes requested):
Hey! 👋 This is Rob's AI reviewer. Thanks for the fix — caught this on the daily grind too.

Clean approach overall. Trying the new `Antigravity IDE` path first, falling back to the old `Antigravity` path if it's missing or unreadable, plus accepting both LS markers — all sensible and minimally invasive. New test coverage for the three cases (v2 hits, v2 missing → v1, v2 corrupt → v1) is solid.

One small thing to fix before merge:

1. **Update `docs/providers/antigravity.md`** — the "Local SQLite Database" section still lists only the old path:
   > **Path:** `~/Library/Application Support/Antigravity/User/globalStorage/state.vscdb`

   Could you add the v2 path and note the fallback order? Something like:

   > **Paths (tried in order):**
   > - `~/Library/Application Support/Antigravity IDE/User/globalStorage/state.vscdb` (v2 — renamed app)
   > - `~/Library/Application Support/Antigravity/User/globalStorage/state.vscdb` (v1 — pre-rename, retained for backwards compat)

   Also, the `Requires:` line at the top of the page still says "Antigravity IDE running" which is fine, but mentioning that pre-rename installs of "Antigravity" also work would be helpful for users not yet on the v2 build.

Everything else (plugin.js, plugin.json, tests, marker logic) looks good. I tested it locally against my own Antigravity IDE install and the v2 path + OAuth refresh + Cloud Code flow works end-to-end.

sakn0m:
Hey @kvanzuijlen, I tested your PR and it works well — I've opened #509 which includes your changes plus two additions:

1. A fix for language server discovery on v2.0.6 (the binary name changed from `language_server_macos` to just `language_server`)
2. Segmented progress bars to match Antigravity's 5-step quota granularity

If you'd prefer to add these to your own branch instead, let me know.

kvanzuijlen:
Hi @sakn0m,

I think it makes sense to keep them separate. I'll add the changes requested by @robinebers. Your branch seems to contain some more changes (also to the UI) and changes to the process name, which I think the maintainers would like to test separately.

kvanzuijlen:
Hi @robinebers,

Good catch on the docs. I also noticed the models are outdated in the docs, but haven't gotten to figuring out how I can find the internal IDs, so decided to keep it out of this PR for now.

robinebers:
Closing as superseded by #538, which consolidates the Antigravity/agy work into one replacement PR and credits this PR's contributor in the body.

PR #511, opened 2026-05-25
Fall back to Extra usage spent in Claude tray title

benmaliel-sc:
## Summary
- Claude Enterprise accounts only enforce the monthly `extra_usage` $ budget — the API returns no Session/Weekly limits, so the tray title showed `--%` (provider style) or was blanked (donut/bars) even though the popup correctly listed the spend.
- Add `primaryOrder` to Weekly (2) and Extra usage spent (3) in `plugins/claude/plugin.json`. The tray selector already falls through to the first candidate present in live data, so a $-budget-only response now resolves to the Extra usage bar.
- Side benefit: also fixes a latent gap where a Weekly-only response would have blanked the title.

## Before / After

Tray title now reads `92%` (was `--%`) for an Enterprise account whose only enforced limit is the monthly `extra_usage` $ budget. Verified across all three icon styles (provider, donut, bars).

Claude popup — Extra usage spent driving the title
Settings — Menubar Icon style options
Tray title — provider style

## Test plan
- [x] `bun run test` — added two regressions (plugin probe emits `extra_usage` as a progress line with finite limit; tray selector resolves Session → Weekly → Extra usage spent chain).
- [x] `bun run build` (tsc + vite) clean.
- [x] Verified live in `bun tauri dev` on a Claude Enterprise account — tray title shows `92%` (was `--%`) across provider, donut, and bars styles.
- [x] Pro/Max accounts unchanged by design (`Session` still wins because `primaryOrder: 1` < `2` < `3`).

---
## Summary by cubic
Fixes the Claude tray title so Enterprise accounts with only a monthly $ budget show a percentage instead of --%, across provider, donut, and bars. Falls back to “Extra usage spent” when Session/Weekly limits are missing.

- **Bug Fixes**
  - Set `primaryOrder` for “Weekly” (2) and “Extra usage spent” (3) in `plugins/claude/plugin.json` to enable Session → Weekly → Extra usage spent fallback (also covers Weekly-only cases).
  - Emit “Extra usage spent” as a progress line with finite used/limit in dollars.
  - Added tests for plugin probe and tray primary bar fallback; Pro/Max behavior unchanged (Session still preferred).

<sup>Written for commit 2d48fc7b56b217dcb61d9d6b024342a1cce5f98c. Summary will update on new commits. <a href="https://cubic.dev/pr/robinebers/openusage/pull/511?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/plugins/claude/plugin.json
+++ b/plugins/claude/plugin.json
@@ -12,10 +12,10 @@
   ],
   "lines": [
     { "type": "progress", "label": "Session", "scope": "overview", "primaryOrder": 1 },
-    { "type": "progress", "label": "Weekly", "scope": "overview" },
+    { "type": "progress", "label": "Weekly", "scope": "overview", "primaryOrder": 2 },
     { "type": "progress", "label": "Sonnet", "scope": "detail" },
     { "type": "progress", "label": "Claude Design", "scope": "detail" },
-    { "type": "progress", "label": "Extra usage spent", "scope": "detail" },
+    { "type": "progress", "label": "Extra usage spent", "scope": "detail", "primaryOrder": 3 },
     { "type": "text", "label": "Today", "scope": "detail" },
     { "type": "text", "label": "Yesterday", "scope": "detail" },
     { "type": "text", "label": "Last 30 Days", "scope": "detail" }
--- a/plugins/claude/plugin.test.js
+++ b/plugins/claude/plugin.test.js
@@ -583,6 +583,31 @@ describe("claude plugin", () => {
     expect(result.lines.find((line) => line.label === "Extra usage spent")).toBeTruthy()
   })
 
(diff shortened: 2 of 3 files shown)


robinebers:
That is a very good addition, and I appreciate it. I never really think about enterprise customers. It is one of these things that gets put on the back burner due to the plugin system, which we are looking to change. 

In the meantime, I'll take a look at this for the next version. Thank you!

benmaliel-sc:
Thanks for the feedback Robin! This is actually my first ever contribution
to an open source project :) I really like the product!

I’m using at work and a few of my peers are too!

robinebers:
Good news! I just merged another PR that was more surgical that will solve this, and I should release this today.

PR #514, opened 2026-05-27
[antigravity] Add keychain auth for agy CLI; stop bundle overwriting user plugins on restart

yesvus:
## Summary by cubic
Adds keychain-based auth for the Antigravity/Gemini CLI so the plugin can use tokens from the OS keychain when LS/DB are unavailable. Also prevents app restarts from overwriting user-installed plugins.

- **New Features**
  - Reads tokens from OS keychain using service names: gemini, gemini-cli-oauth, antigravity-cli-oauth.
  - Supports go-keyring base64 envelope and common JSON shapes (access_token/refresh_token, nested tokens, expiry fields).
  - Prefers DB tokens; uses non-expired keychain access tokens or refreshes via keychain refresh token when needed.

- **Bug Fixes**
  - Skip copying plugin files if the destination already exists in `src-tauri` to avoid overwriting user plugins.
  - Guard keychain expiry parsing against NaN to avoid invalid token handling.

<sup>Written for commit 8c5e8394bfd7fd4de3e4274bf77002a34ffb9236. Summary will update on new commits. <a href="https://cubic.dev/pr/robinebers/openusage/pull/514?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/plugins/antigravity/plugin.js
+++ b/plugins/antigravity/plugin.js
@@ -548,7 +548,7 @@
       }
     }
 
-    ctx.host.log.info("no antigravity tokens found in keychain")
+    ctx.host.log.debug("no antigravity tokens found in keychain")
     return null
   }
 
--- a/plugins/antigravity/plugin.js
+++ b/plugins/antigravity/plugin.js
@@ -457,10 +457,106 @@
     return { plan: plan, lines: lines }
   }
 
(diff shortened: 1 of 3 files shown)


robinebers:
Closing as superseded by #538 for agy keychain auth. Gemini CLI fallback services are intentionally excluded, and the bundled-plugin overwrite change is unrelated enough to keep out of the Antigravity replacement PR.

PR #638, opened 2026-06-17
feat(codex): show rate limit resets

### LINKED ISSUE #597: Pull in rate limit resets
> **Migrated from** [robinebers/OpenUsage-Swift#40](https://github.com/robinebers/OpenUsage-Swift/issues/40)
> Issue numbers like `#N` in the text below refer to the [OpenUsage-Swift issue tracker](https://github.com/robinebers/OpenUsage-Swift/issues) unless updated.

See https://github.com/robinebers/openusage/pull/577

robinebers:
## What Changed

Ports the on-demand **rate-limit reset credits** feature from the JS/Tauri edition (merged in [#577](https://github.com/robinebers/openusage/pull/577)) into the native Swift edition.

- `CodexUsageMapper` reads `rate_limit_reset_credits.available_count` from the `wham/usage` response and appends a **Rate Limit Resets** row (e.g. `1 available`) **before** Credits. Malformed/`null` counts are skipped via `ProviderParse.number` (matches the JS guard).
- Exposed as a pinnable menu-bar widget tile (`codex.rateLimitResets`), available in the Add-Widget gallery — not placed/pinned by default, matching how `codex.credits` behaves.

## Implementation note (value plumbing)

The metric is a **verbatim `.text` row** backed by a new `verbatimCount` widget descriptor — the count-flavored twin of the existing `verbatimDollars` used by Credits. This keeps every surface resolving from one `WidgetData`:

- popover / dashboard row → `1 available`
- menu-bar tile → `1` (compact)

An earlier revision modeled this as a `.badge`, which carries only a display string and left `used = 0`, so a pinned tile read `0` while the popover read `1 available`. Switching to the `verbatimCount`/`.text` path (which sets the numeric `used` *and* preserves the raw text) fixed that divergence.

The deeper smell that made the bug possible — formatting a number into a string in the mapper, then re-parsing it back out for the tile — is pre-existing (Credits does the same) and tracked as follow-up tech-debt in **#641**.

## Why

Codex grants on-demand rate-limit resets, but the Swift edition ignored the new usage API field. Tracks issue #597 (migrated from the Swift tracker).

## Verification

- `swift build` — clean.
- `swift test` — 248 passed, 1 pre-existing skip, 0 failures.
  - `CodexUsageMapperTests`: resets shown before Credits (`1 available`); zero count (`0 available`); malformed/`null` omitted.
  - `WidgetDataStoreTests`: regression test asserting the tile resolves `displayedValue == 1` (tray) **and** `unboundedDetail == "1 available"` (popover) from one `WidgetData`.
- Manually verified against the live API: pinned tile reads `1`, popover row reads `1 available`.

Closes #597

### DIFF
--- a/Sources/OpenUsage/Models/WidgetDescriptor+Factories.swift
+++ b/Sources/OpenUsage/Models/WidgetDescriptor+Factories.swift
@@ -91,6 +91,21 @@ extension WidgetDescriptor {
         return make(id: id, provider: provider, metricLabel: metricLabel ?? title, sample: sample)
     }
 
+    /// Unbounded count rendered verbatim from the provider's `.text` line (e.g. Codex rate-limit
+    /// resets "1 available"); the parsed count still feeds the menu-bar tile's compact value, so the
+    /// tray and the popover never diverge.
+    static func verbatimCount(
+        id: String,
+        provider: Provider,
+        title: String,
+        metricLabel: String? = nil
+    ) -> WidgetDescriptor {
+        var sample = WidgetData(title: title, icon: provider.icon,
+                                kind: .count, used: 0, limit: nil)
+        sample.preservesRawText = true
+        return make(id: id, provider: provider, metricLabel: metricLabel ?? title, sample: sample)
+    }
+
     /// Unbounded count resolved from a provider `.badge` line via `valueTextOverride`
     /// (e.g. Grok pay-as-you-go).
     static func badge(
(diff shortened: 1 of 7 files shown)


robinebers:
Follow-up tech-debt tracked in #641: this PR routes rate-limit-resets through the same value path as Credits (which works and matches the current codebase), but the underlying smell — formatting a number into a string in the mapper, then re-parsing it back out for the menu-bar tile — is left for #641 to fix properly (carry the raw number on the row).

PR #668, opened 2026-06-21
fix(dashboard): flag stale snapshots with an "Updated X ago" hint (#582)

### LINKED ISSUE #582: Snapshot cache can show stale plan/limits indefinitely — investigate and document
> **Migrated from** [robinebers/OpenUsage-Swift#11](https://github.com/robinebers/OpenUsage-Swift/issues/11)
> Issue numbers like `#N` in the text below refer to the [OpenUsage-Swift issue tracker](https://github.com/robinebers/OpenUsage-Swift/issues) unless updated.

## Problem

`ProviderSnapshotCache` persists the last successful snapshot in `UserDefaults` and displays it at launch even when expired (stale-while-revalidate). A failed refresh keeps the last good snapshot on screen with only the small amber warning triangle in the section header as a clue.

Observed in practice: after a Claude plan change, the dashboard kept showing the old plan ("Team 5x") and its limits while the credentials and API already returned the new plan ("Max 20x"). The fossil survived across app sessions until a refresh finally succeeded and overwrote it. Why the refreshes failed long enough for that to happen is unknown — the app has no logging, so earlier sessions left no trace.

## Tasks

- [ ] Figure out why/when refreshes fail silently long enough for cached snapshots to fossilize (add minimal logging if needed to diagnose)
- [ ] Document the caching behavior properly (docs/refreshing.md + AGENTS.md already touch it, but the "stale data can persist indefinitely" consequence is not obvious)
- [ ] Make staleness obvious in the UI — e.g. a visible "last updated X ago" hint when a snapshot is older than the refresh interval, or expire the plan badge

## References

- `Sources/OpenUsage/Stores/ProviderSnapshotCache.swift`
- `WidgetDataStore.refresh` stale-friendly error path

davidarny:
Fixes #582.

## Summary

A failed refresh keeps the last good snapshot on screen, which is the intended stale-while-revalidate behavior. The gap: when refreshes keep failing, the old plan and limits stay up for hours with only a small warning triangle to hint at it, so a plan that changed on the provider's side reads as current.

This adds an "Updated X ago" hint next to the provider's name once its snapshot ages past its freshness window, so fossilized data stops passing for live.

## Root cause

`WidgetDataStore.refresh` preserves `snapshots[providerID]` on a failed pass and raises `providerErrors` instead. The header showed the plan badge and the error triangle, but nothing about how old the displayed numbers were, so a snapshot could persist across sessions until a refresh finally succeeded.

## What changed

- `WidgetDataStore.stalenessLabel(for:)` returns `"Updated X ago"` once a snapshot is older than `stalenessThreshold` (twice the refresh interval). Healthy providers reset their age to ~0 each pass and only brush one interval before the next, so two intervals keeps the hint off them.
- `ProviderSectionHeader` renders the hint as muted tertiary text after the plan badge, hidden while a refresh is in flight (the spinner already covers that). The provider name keeps its width so a long age string truncates instead of wrapping the header.

- `docs/refreshing.md` gains a "Stale data" section.

## Screenshots

CleanShot 2026-06-21 at 14 31 32@2x

CleanShot 2026-06-21 at 14 32 22@2x

## Testing

`swift test` passes (333 tests). The new `StalenessLabelTests` covers the threshold boundary (`>=` at exactly two intervals), day-level formatting, a future timestamp from clock skew, and the end-to-end failing-refresh case that keeps the fossil on screen but labels it.

Checked in the dev build by backdating the cached snapshots and forcing refreshes to fail through a dead proxy. The header showed "Updated 3h 2m ago" with the warning triangle on both Claude and Codex while the old numbers stayed put.


robinebers:
Super cool, and something I've been thinking about as well. The updated label only shows up when it's stale or always?

Haven't tested it yet, but I would love it to be only visible after it becomes stale. 

I also think that because some plans like super Grok heavy are very long, we don't need to show when it was last updated. Instead, just say "outdated" with the warning triangle or something.

What do you think?

image

davidarny:
@robinebers 

> The updated label only shows up when it's stale or always?

Yes, it shows only when stale. It's not showing always. While the snapshot is fresh `stalenessLabel` returns `nil` and the header renders nothing, so there's no permanent label.

UPD: Changing "Updated" label to just "Outdated"

robinebers:
> @robinebers
> 
> > The updated label only shows up when it's stale or always?
> 
> Yes, it shows only when stale. It's not showing always. While the snapshot is fresh `stalenessLabel` returns `nil` and the header renders nothing, so there's no permanent label.

That's great, thank you. Just what do you think about the label, because as you can see, Super Grok Heavy is quite long, and that would mean that the label either gets cut off or, worse, pushes it out or moves into this next row. 

could be shorter maybe like

- `Outdated ⚠️`
- `Update falied ⚠️` 

and then use the instant tooltips we have to add context 

> "Last updated failed 3h 12m ago."

davidarny:
@robinebers here how it looks now

CleanShot 2026-06-21 at 15 28 33@2x

robinebers:
This looks awesome, thank you, man.

PR #672, opened 2026-06-21
perf(refresh): cache ccusage runner resolution and snapshot blob

davidarny:
## What

Two refresh-loop hot paths did avoidable work on every pass. Both fixes are internal; behavior is unchanged.

### Runner resolution (`CcusageRunner`)

`query()` resolved all five package runners up front, spawning a `--version` probe subprocess for each one not found at a hardcoded path, then ran ccusage with only the first. The lower-priority probes were pure waste, repeated on every cache-miss for Claude and Codex.

It now resolves lazily (stops at the first runner that runs ccusage) and caches the winning runner for the session, so later passes skip resolution. A failed cached runner is dropped and not re-run in the same pass.

| | before | after |
|---|---|---|
| spawns / query | 5 | **1** |
| spawns / Claude+Codex pass | 10 | **2** |

### Snapshot cache (`ProviderSnapshotCache`)

Every `snapshot()` / `store()` re-decoded the entire all-providers JSON blob from `UserDefaults`, so one refresh pass paid O(N) full decodes plus O(N) encodes on the MainActor. The cache now keeps an in-memory mirror and writes through, decoding at most once per instance.

Measured over the real type, 200 passes × 5 providers, release build:

| | before | after | Δ |
|---|---|---|---|
| CPU time | 0.267 s | **0.083 s** | **−69%** |
| instructions retired | 6.07 B | **1.88 B** | **−69%** |

A Time Profiler trace confirms the decode frames (`JSONDecoder`, `JSONScanner`, `MetricLine.init(from:)`) drop to zero; only the write-through encode remains.

## How

- Lazy, session-cached runner resolution in `query()`.
- In-memory write-through mirror for the snapshot blob.
- Both backed by `OSAllocatedUnfairLock`.

## Tests

New regression guards: spawn counts and cached-runner fallback in `CcusageRunnerTests`, cache write-through and persistence in `ProviderSnapshotCacheTests`. All 329 tests pass.

## Verified live

Built and launched the optimized app (`script/build_and_run.sh run`, release) and watched the log through a real refresh of both ccusage providers:

```
[cache]          loaded 2 snapshots from disk
[plugin:ccusage] launch bunx claude daily   →  resolved bunx
[plugin:ccusage] launch bunx codex daily    →  resolved bunx
[cache]          write codex
[refresh]        codex ok (2393ms)
[cache]          write claude
[refresh]        claude ok (2480ms)
[refresh]        batch end (2526ms, 2 ok / 0 failed)
```

The pass spawned three subprocesses total: one keychain read plus the two `bunx` ccusage queries. No `--version` probe spawns and no pnpm/yarn/npm/npx launches, so lazy resolution stopped at bunx instead of probing the four runners it would never use. The cache decoded the blob once at launch (`loaded 2 snapshots from disk`) and wrote through on each refresh, with no re-decode in between.

### DIFF
--- a/Sources/OpenUsage/Services/CcusageRunner.swift
+++ b/Sources/OpenUsage/Services/CcusageRunner.swift
@@ -1,4 +1,5 @@
 import Foundation
+import os
 
 enum CcusageProvider: String, Sendable {
     case claude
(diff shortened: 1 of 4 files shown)


validatedev:
@codex review

robinebers:
Beautiful, thanks David!

PR #677, opened 2026-06-22
feat(dashboard): add provider-header drag handle

robinebers:
## Summary
- Lead each dashboard provider header with a quiet 2×3 dot grip so the drag-to-reorder affordance is discoverable. The whole header line still carries the gesture; the grip is purely visual (open-hand cursor was tried and removed, the grip is enough).
- Mirror the grip in the lifted reorder preview so the floating chip matches the live header.
- Nudge the grip a touch left (smaller left inset) and keep a small gap to the name; Customize keeps its symmetric inset so its name doesn't shift.
- Bump the provider-name point size slightly so the section title leads its group.
- Tighten the top padding between stacked text-only rows (Today / Yesterday / Last 30 Days) so runs of one-liners read as one cluster.
- chore: add `node_modules/` to `.gitignore`.

## Screenshots

image

## Test plan
- [ ] Build and run the menu-bar app (`./script/build_and_run.sh verify`).
- [ ] Open the dashboard popover and confirm the dot grip sits to the left of each provider name.
- [ ] Drag a provider header to reorder providers; confirm the floating preview shows the same grip.
- [ ] Confirm the Customize screen's provider names are unchanged (no grip, same inset).
- [ ] Check Compact and Regular densities both look right.

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,7 @@
 .DS_Store
 .build/
 dist/
+node_modules/
 *.log
 .agents/worktrees/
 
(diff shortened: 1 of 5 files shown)


davidarny (approved):
LGTM!

PR #720, opened 2026-06-24
[codex] Add provider-aware refresh scheduling

### LINKED ISSUE #676: Claude rate limited, retry in ~xxm
### Description

Hello, I've already logged in again to Claude but it still not able to fetch rates. It happened just suddenly without any notifications. It's happening for the past 5 days already.

Image

### Steps to Reproduce

1. Open OpenUsage and view Claude section

### Expected Behavior

Should not be rate-limited

### Actual Behavior

It is being rate-limited

### OpenUsage Version

0.6.27

### Operating System

macOS (Apple Silicon)

### Screenshots

In the main description above

### Additional Context

_No response_

### LINKED ISSUE #678: Polling Interval
### Problem

Hi, 

thank you for the great product in OpenUsage, I am having an amazing time using it. Its especially great to see how stable it works compared to alternatives. 

I would request a 1 minute polling time. For Codex if you hit your 5 hour usage limit, the normally not rate limited request that you can run through the web ui also get rate limited, so its very important for me to stay above 0%. 

I don t see an issue with polling more regularly, but you probably  now better, let me know! Thanks for your work.

Greetings from Vienna,
Ferdi

### Proposed Solution

Poll Every Minute

### Alternatives Considered

Let the User define the polling interval, but i suspect there is some reasoning behind having a polling time, that users might not be aware of. 

### Additional Context

_No response_

### LINKED ISSUE #679: Smart / adaptive refresh timer with per-provider cadence and exponential backoff
### Problem

We currently use a strict global 5-minute refresh interval for all providers (see `RefreshSetting.defaultMinutes = 5` and the periodic loop in `AppContainer.startPeriodicRefresh` / `waitForNextRefresh`).

A simple per-provider 60-second `failureRetryBackoff` exists in `WidgetDataStore` to protect against wake storms after a failure, but it is not exponential, not per-provider tuned, and the main cadence remains fixed at 5 minutes for successes.

Some providers are very sensitive:
- Grok is known to block rapid polling quickly.
- Claude's usage API also rate-limits (users see "Rate limited, retry in ~Nm" and there are existing reports).

A fixed aggressive interval risks bans / throttling. A fixed conservative interval (5 min) can be too slow for users who want fresher data (see #678 requesting 1 min for Codex near session limits).

### Goal

Introduce a "smart" or "intelligent" update timer:

- **Per-provider base / default interval** (good starting point chosen per provider instead of one global).
- **Simple exponential backoff on consecutive failures**, with a reasonable ceiling (around 15 minutes).
- On success, return promptly to the provider's normal base cadence.
- Force refreshes (⌘R / footer) and re-enables must still bypass backoff/caches as today.

### Example (Codex, illustrative)

Base: 60 seconds

Failure sequence:
- 1st failure → next probe after ~2 minutes
- 2nd failure → ~5 minutes
- 3rd failure → ~10 minutes
- Subsequent failures → cap at ~15 minutes

Success at any point resets the provider back to its base interval.

### Proposed Behavior / Scope

- Move away from a single global `RefreshSetting.interval` toward provider-aware scheduling.
- The background loop should support independent per-provider next-probe times (or compute the next global wake as the soonest pending provider).
- Generalize / replace the fixed `failureRetryBackoff` with a simple per-provider exponential backoff state machine.
- Snapshot cache TTL should align with the effective interval for that provider (or stay conservative).
- Staleness hints and "Next update in" UI should continue to make sense (may need adjustment).
- Keep everything simple and observable; avoid complex user-facing controls initially.
- Update `docs/refreshing.md` and any architecture notes.

### Related

- #678 Polling Interval (user request for shorter cadence on Codex)
- #676 Claude rate limited reports

### Additional Context

We introduced the 5-minute floor precisely to avoid spamming providers. A smart backoff gives us the best of both worlds: fresher data for well-behaved providers while backing off quickly and gracefully for the ones that push back.

### LINKED ISSUE #680: Establish per-provider base refresh intervals and document provider rate limit behavior
### Problem

The current refresh system is a single hardcoded 5-minute global cadence (`RefreshSetting`). As we move toward a smart / adaptive timer (#679), we need to decide what the "normal" (success path) cadence should be for each provider.

Different providers have different characteristics and tolerance:

- Codex (ccusage + chatgpt backend): Users sometimes want faster updates when near session limits (see #678). Base suggestion in discussion: ~60s.
- Claude (Anthropic usage API via ccusage): Already documents rate limiting ("Rate limited, retry in ~Nm"). Need a safe but useful default.
- Grok: Explicitly called out as blocking rapid requests quickly. Likely needs a longer or more conservative base.
- Cursor, Devin: TBD.

We also lack consolidated knowledge of:
- What the provider's own rate limits / recommended polling windows are.
- Whether providers return `Retry-After` or similar headers we could honor.
- Real-world failure modes when polling too fast.

### Goal

- Choose and document a sensible **base (success) refresh interval per provider**.
- Update provider docs (under `docs/providers/`) and the main refreshing doc with any known rate limit / backoff guidance.
- Possibly add light runtime telemetry or logging when we hit throttling responses so we can tune later.
- Make the chosen bases easy to adjust in code (e.g. a per-provider value in the `Provider` model, the runtime, or a small table) rather than magic numbers scattered around.

### Non-goals (for now)

- Exposing a user-facing per-provider interval slider (the smart backoff + good defaults should be enough).
- Complex per-endpoint or per-metric scheduling.

### Related

- #679 Smart / adaptive refresh timer with per-provider cadence and exponential backoff
- #678 Polling Interval
- #676 Claude rate limited

### Acceptance ideas

- Every provider has a clear, reviewable base interval constant or property.
- `docs/refreshing.md` (and provider pages) explain the cadence strategy and what to expect on throttling.
- No change in observed behavior for providers until the smart timer lands (or the two can be done together).

### LINKED ISSUE #681: Make 'Next update' countdown and staleness hints aware of per-provider / backoff schedules
### Context

Today the footer shows "Next update in Nm" and staleness detection ("Outdated" after ~10 min) are driven by the global fixed `RefreshSetting.interval` (5 minutes) and `WidgetDataStore.stalenessThreshold = interval * 2`.

See:
- `AppContainer.waitForNextRefresh` (sleeps the global interval)
- `lastRefreshAt` + footer logic
- `stalenessThreshold` and `stalenessHint(for:)`
- `ProviderSnapshotCache` TTL tied to the interval

Once we have a smart timer (#679) with per-provider bases and variable exponential backoffs, these will be wrong or misleading:
- Different providers will legitimately have different "next probe" times.
- After failures, a provider may legitimately wait up to ~15 min before the next attempt.
- The dashboard may show a global countdown that doesn't match reality for any given provider.

### Task

- Track per-provider "next expected probe time" (base cadence or current backoff deadline).
- Update the "Next update in ..." display to reflect the soonest actionable next probe across enabled providers (or make it per-section if that feels better).
- Adjust staleness detection so it only flags real staleness (data older than N effective cycles for that provider) rather than a global wall time.
- Ensure the snapshot cache TTL logic doesn't fight the new schedule.
- Keep the existing behavior for the simple fixed case during the transition.

### Related

- #679 Smart / adaptive refresh timer with per-provider cadence and exponential backoff
- #680 Establish per-provider base refresh intervals and document provider rate limit behavior

robinebers:
## Draft note

This PR changes the default refresh rates by provider. 

- Codex: 1 min
- Claude: 3 min
- Grok/Devin/Cursor: 5 min

This works, but the problem is that the timer will now always say "next update in one minute" (= Codex). We need a solution for this. 

## Summary

- Adds smarter refresh timing per provider.
- Backs off repeated failures automatically.
- Keeps manual refresh immediate.

Closes #676
Closes #678
Closes #679
Closes #680
Closes #681

### DIFF
--- a/Sources/OpenUsage/App/AppContainer.swift
+++ b/Sources/OpenUsage/App/AppContainer.swift
@@ -57,33 +57,34 @@ final class AppContainer {
 
     deinit { refreshTask.cancel() }
 
-    /// Drives live updates: refresh on launch, then again every refresh interval. Each pass honors the
-    /// cache, so it only hits the network once a snapshot has actually expired. `@Observable` propagates
-    /// the resulting snapshot changes to the menu-bar label and any open widgets, so the UI refreshes on
-    /// its own instead of only when the popover opens.
+    /// Drives live updates: refresh due providers on launch, then wake for the soonest provider-specific
+    /// probe. Each pass honors the cache/backoff, so Codex can be fresh every minute without forcing
+    /// Claude/Grok/Cursor/Devin to poll at the same pace.
     private static func startPeriodicRefresh(dataStore: WidgetDataStore) -> Task<Void, Never> {
         Task {
             while !Task.isCancelled {
-                await dataStore.refreshAll()
-                await waitForNextRefresh()
+                await dataStore.refreshDueProviders()
+                await waitForNextRefresh(dataStore: dataStore)
             }
         }
     }
 
-    /// Sleep for the refresh interval, but wake early when the user enables/disables a provider so a
-    /// newly-enabled provider is fetched promptly instead of waiting out the full interval. Each pass still
-    /// honors the cache (and the per-provider failure backoff), so an early wake only hits the network for
-    /// a provider whose snapshot has actually expired.
+    /// Sleep until the soonest provider is due, but wake early when the user enables/disables a provider
+    /// so a newly-enabled provider is fetched promptly instead of waiting out another provider's cadence.
+    /// Each pass still honors the cache and per-provider backoff.
     ///
     /// Deliberately scoped to `ProviderEnablementStore.didChangeNotification` — NOT the firehose
     /// `UserDefaults.didChangeNotification`, which fires for the app's own snapshot-cache writes, Sparkle's
     /// update bookkeeping, and unrelated global-domain changes from other processes. Waking on that, with
     /// no minimum interval before re-refreshing, collapsed the fixed 5-minute cadence into a refresh storm.
-    private static func waitForNextRefresh() async {
-        let interval = RefreshSetting.interval
+    private static func waitForNextRefresh(dataStore: WidgetDataStore) async {
+        let nextRefreshAt = dataStore.nextScheduledRefreshDate()
         await withTaskGroup(of: Void.self) { group in
-            group.addTask {
-                try? await Task.sleep(for: .seconds(interval))
+            if let nextRefreshAt {
+                group.addTask {
+                    let delay = max(0.25, nextRefreshAt.timeIntervalSinceNow)
+                    try? await Task.sleep(for: .seconds(delay))
+                }
             }
             group.addTask {
                 for await _ in NotificationCenter.default.notifications(named: ProviderEnablementStore.didChangeNotification) {
--- a/Sources/OpenUsage/Models/ProviderSnapshot.swift
+++ b/Sources/OpenUsage/Models/ProviderSnapshot.swift
@@ -7,19 +7,24 @@ struct ProviderSnapshot: Hashable, Sendable, Codable {
     var plan: String?
     var lines: [MetricLine]
     var refreshedAt: Date
+    /// Optional provider-supplied deadline for the next live probe, used when an API asks us to slow
+    /// down (for example Claude's `Retry-After` on rate limiting).
+    var retryAfter: Date?
 
     init(
         providerID: String,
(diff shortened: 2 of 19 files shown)


davidarny (approved):

validatedev:
@codex review

robinebers:
This should probably be closed because it runs into issues where we suddenly don't know "when" the next update is (e.g., the next update is always 1 min away if codex becomes 1 minute, which is a lie as the next claude update could be 5 min away)

PR #726, opened 2026-06-25
feat(popover): System Settings grouped surface + Settings-primary footer

robinebers:
## What & why

Two popover refinements: make the data region read like a native macOS System Settings grouped pane, and put the more-used destination up front.

### Grouped surface (semantic, no tuned values)
Provider/settings cards now use macOS's own grouped-box treatment — the bright page tray (`textBackgroundColor`) lifted by **`.fill.quaternary`**, the system's subtle grouped fill (≈`#F9F9F9` over white in light; a step lighter than the page in dark), borderless. It tracks light/dark and Increase Contrast automatically, with **no hand-tuned hex values**. Replaces the previous tuned gray-tray/white-card look (and the short-lived Elevated/System toggle, which is dropped).

Why a semantic fill and not a native grouped `List`: the dashboard's drag-to-reorder is a custom local-gesture system *specifically because* the menu-bar panel can't host a system drag session — which native `List`/`Form` reorder relies on. Matching the surface semantically on the existing structure leaves reorder, scroll, and the screen-slide animation untouched.

### Footer: Settings-primary
The footer split button's primary action is now **Settings**; **Customize** moves into the `⌄` overflow menu. Everyday layout edits (reorder, hide, pin) are already reachable by dragging and right-clicking the dashboard, so Settings is the more frequent deliberate destination. `⌘,` and `⏎` shortcuts are unchanged (handled by the always-on key monitor).

## Notes
- Docs (dashboard, settings) and the surface-opacity test updated to match.
- Modern Liquid Glass controls (the SDK-26 linked-SDK stamp, #725) are already on `swift`; this branch inherits them.

## Test plan
- `swift build` + `swift test` — **436 tests pass** (1 skipped).
- Verified the grouped surface in light + dark and the new footer label in a signed SDK-26 dev build.

---

> [!NOTE]
> **Low Risk**
> UI-only theming and footer navigation reordering; no auth, data, or API changes.
> 
> **Overview**
> The popover **data region** no longer uses hand-tuned gray tray / white card colors and card hairlines. The tray is **`textBackgroundColor`**, and grouped cards are a **borderless** stack of opaque tray plus **`.fill.quaternary`** (System Settings–style grouping that follows light/dark and Increase Contrast). Live and lifted drag previews share that one surface; lift depth stays on shadow.
> 
> The dashboard footer **split button** now leads with **Settings**; **Customize** moves into the **⌄** overflow menu. **⌘,** and **⏎** behavior is unchanged via `PopoverKeyReader`. Docs and `PopoverSurfaceOpacityTests` drop the standalone card `NSColor` opacity check in favor of tray-only assertion plus a comment that cards are opaque by construction.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 963730e1e1f9367e6ea1b1431f3673b8f3d774ed. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/OpenUsage/Support/Theme.swift
+++ b/Sources/OpenUsage/Support/Theme.swift
@@ -5,9 +5,10 @@ import AppKit
 ///
 /// The popover is always a solid, opaque panel — glass is reserved for the footer chrome (its frosted
 /// bar and the controls on it), never the data region behind it (Apple's guidance: glass for
-/// navigation/controls, content on an opaque surface). So the data surfaces here are plain opaque fills: a light-gray
-/// window "tray" with white grouped cards in light mode, a near-black tray with a step-lighter card
-/// in dark mode — the System Settings grouped-box look.
+/// navigation/controls, content on an opaque surface). The data region mirrors the macOS System
+/// Settings grouped look: a bright page "tray" with borderless grouped cards lifted off it by the
+/// system's own `.fill.quaternary` (the same subtle fill Settings' grouped boxes use — no hand-tuned
+/// values), so it adapts to light/dark like every other Mac app.
 enum Theme {
     /// Hierarchical secondary tint for the provider marks.
     static let iconGray = AnyShapeStyle(.secondary)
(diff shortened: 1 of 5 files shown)


robinebers:
Triage of the Cursor Bugbot note (HeaderView.swift:106, low severity):

Accurate but expected, not a regression worth a fix. `⌘,` is inert **only while the `⌄` overflow menu is open** — while an `NSMenu` is tracking, the panel isn't key, so the always-on `PopoverKeyReader` monitor pauses (standard macOS behavior: global shortcuts don't fire during menu tracking). The same is true of `⏎` for Customize, so it's symmetric. In every normal state (popover focused, menu closed) `⌘,` opens Settings via the monitor, and Settings is now the always-visible primary button right next to the chevron — so there's no practical loss.

Registering `⌘,` on the button wouldn't help (the button's shortcut also can't fire while the menu owns key), and re-adding a redundant "Settings" item to a menu whose sibling button already is Settings would be odd UX. Leaving as-is.

PR #739, opened 2026-06-25
feat(telemetry): anonymous crash + uncaught-exception reporting

robinebers:
## What & why

Telemetry (PR #735) reported only coarse provider-refresh error *counts* — app crashes and uncaught exceptions went unreported. This captures them, anonymously, gated by the **same** Settings → Privacy → *Share Anonymous Usage* opt-out.

**Backend decision:** PostHog Error Tracking, not Sentry (owner call). It reuses the existing posthog-ios SDK/token/dashboard and the opt-out plumbing, so it adds **no new shipped dependency**. Trade-off accepted: PostHog labels Swift traps as bare `SIGTRAP` and lacks Sentry's crash-triage tooling; in exchange we keep one vendor and one toggle. (The engine underneath is vendored PLCrashReporter — Mach + uncaught-exception handlers — so native capture itself is mature.)

## Changes
- **`Telemetry.swift`** — `config.errorTrackingConfig.autoCapture = enabled`. Gating *install* (not just sending) on the opt-out means an opted-out launch installs no handler and writes nothing to disk; `optOut()` (mirrored by `setEnabled`) also uninstalls the integration and ignores captures at runtime. Anonymous: no `identify`, `personProfiles = .never`.
- **`release.sh`** — build with `-Xswiftc -g` (DWARF; **zero** binary-size cost — DWARF stays in `.o`, extracted into the dSYM), then `dsymutil` **after** the vtool SDK restamp, with a guard that every shipped-binary UUID is present in the dSYM.
- **`release.yml`** — upload the dSYM via `posthog-cli dsym upload --directory dist/dSYMs`. Skips with a loud warning if `POSTHOG_CLI_API_KEY` is unset, so a missing key never blocks a release.
- **`docs/privacy.md`** — crash reporting honors the same toggle; no PII (stack traces of our own code only).
- **`TelemetrySinkTests.swift`** — regression test pinning *opt-out ⇒ no crash autocapture*.

## Dependency justification (per AGENTS.md)
No new **shipped** dependency — PostHog is already in `Package.swift`. The only addition is `@posthog/cli`, a **CI-only** npm tool used in the release workflow to upload dSYMs; it is never embedded in the app bundle.

## Verification
- Full suite: **444 pass, 0 fail** (incl. the new opt-out test).
- On the real app binary: `vtool` restamp **preserves `LC_UUID`** (identical before → after → in the dSYM), and `-g`+`dsymutil` yields **61,944 DWARF line-table rows** — so the Liquid Glass restamp does **not** break symbol upload.
- Inert/no-token path confirmed to make **zero** PostHog network calls.
- SDK source confirms opt-out uninstalls the crash integration + ignores captures.

## Action items before crashes symbolicate
- [x] "Exception autocapture" enabled in the PostHog project (already on; it's a server-side gate the SDK reads from cache, so capture arms on the *second* launch after enabling).
- [ ] Add GitHub secrets **`POSTHOG_CLI_API_KEY`** (scopes: error-tracking write + org read) and **`POSTHOG_CLI_PROJECT_ID`**. Without them, the upload step skips (warns) and crashes won't symbolicate.

## Note
Dev builds (`com.robinebers.openusage.dev`) use the same baked token, so dev crashes also report — tagged `app_version = 0.7.0-dev`, unsymbolicated. Consistent with the existing usage-telemetry posture; excluding dev would be a uniform follow-up across all telemetry.

---

> [!NOTE]
> **Medium Risk**
> Changes privacy-sensitive telemetry behavior and the release workflow; mitigated by explicit opt-out gating, docs, and CI guards on dSYM presence and upload failures when secrets are set.
> 
> **Overview**
> Adds **anonymous crash and uncaught-exception reporting** through the existing PostHog SDK, using the same **Share Anonymous Usage** opt-out as daily usage telemetry. `PostHogTelemetrySink` turns on `errorTrackingConfig.autoCapture` only when telemetry is enabled at init (via a testable `errorAutocaptureEnabled` helper), so an opted-out launch does not install crash handlers or write reports locally.
> 
> The **release pipeline** now emits symbol-friendly builds and uploads matching debug symbols: `release.sh` builds with `-Xswiftc -g`, runs `dsymutil` on the post-`vtool` shipped binary, and fails if any Mach-O UUID is missing from the dSYM. **release.yml** adds a PostHog dSYM upload step (`posthog-cli`); it warns and skips when `POSTHOG_CLI_API_KEY` is unset but fails the release if upload fails when the key is present.
> 
> **docs/privacy.md** documents crash reports under the same toggle, and **TelemetrySinkTests** locks the opt-out contract.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 284fa8472745901af9850cc13f8250f34158f4d3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -115,6 +115,29 @@ jobs:
           OPENUSAGE_VERSION: ${{ steps.meta.outputs.version }}
         run: ./script/release.sh
 
+      # Upload the dSYM produced by release.sh so PostHog can symbolicate crash reports (the in-app
+      # crash autocapture is anonymous and honors the Share-Anonymous-Usage opt-out; see docs/privacy.md).
+      # posthog-cli is a CI-only build tool (npm), never shipped in the app bundle. Skips with a loud
+      # warning when POSTHOG_CLI_API_KEY is unset (e.g. forks), so a missing key never blocks a release;
+      # when the key IS set, an upload failure fails the step so we never silently ship unsymbolicated.
+      - name: Upload dSYMs to PostHog
+        env:
+          POSTHOG_CLI_API_KEY: ${{ secrets.POSTHOG_CLI_API_KEY }}
+          POSTHOG_CLI_PROJECT_ID: ${{ secrets.POSTHOG_CLI_PROJECT_ID }}
+          # US region, matching TelemetryConfig.host in the app.
+          POSTHOG_CLI_HOST: https://us.i.posthog.com
+          VERSION: ${{ steps.meta.outputs.version }}
+        run: |
+          if [ -z "${POSTHOG_CLI_API_KEY}" ]; then
+            echo "::warning::POSTHOG_CLI_API_KEY not set — skipping dSYM upload; crashes for $VERSION will NOT symbolicate."
+            exit 0
+          fi
+          [ -d dist/dSYMs ] || { echo "dist/dSYMs missing — release.sh did not produce a dSYM." >&2; exit 1; }
+          npm install -g @posthog/cli
+          # Symbolication matches a crash report to its symbols by the binary's Mach-O UUID (carried in
+          # the dSYM), so `--directory` is all that's required — no per-release flags to keep in sync.
+          posthog-cli dsym upload --directory dist/dSYMs
+
       - name: Publish DMG to GitHub Releases
         uses: softprops/action-gh-release@v3
         with:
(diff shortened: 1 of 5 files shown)


robinebers:
bugbot run

PR #751, opened 2026-06-26
fix(claude): restore the weekly Sonnet limit

validatedev:
## What

Re-adds the separate weekly **Sonnet** limit to the Claude provider. Anthropic has brought the limit back upstream — the `seven_day_sonnet` field is present in the usage API response again (the Anthropic console once more shows distinct "All models" and "Sonnet" weekly meters), so the metric should reappear.

This reverses #744, which dropped the metric when the field disappeared.

## Changes

Restored end to end, matching the metric's original defaults:

- **Widget descriptor** (`ClaudeProvider`) — `claude.sonnet`, ordered after Weekly.
- **Mapper** (`ClaudeUsageMapper`) — reads `seven_day_sonnet` as a weekly-period window.
- **Default layout** (`DefaultLayout`) — off by default, member of `expandedMetricIDs` so it sits below the per-provider caret if a user enables it.
- **Tests** — `ClaudeUsageMapperTests` and `LayoutStoreTests` assertions for the Sonnet row.
- **Docs** — provider reference, dashboard guide, README.

## Notes

The mapper reads each usage window behind a presence guard (`guard let object = value as? [String: Any]`), so responses that omit `seven_day_sonnet` simply skip the row rather than showing an empty meter. This keeps the change backward-compatible with any account/plan that doesn't expose a separate Sonnet limit.

## Verification

- `swift build` — succeeds (pre-existing unrelated warnings only).
- `swift test --filter 'ClaudeUsageMapperTests|LayoutStoreTests'` — 45 tests pass, including the restored Sonnet assertions.

<h3>Confidence Score: 5/5</h3>

Safe to merge — the change is a clean, contained restoration of a single optional metric with full backward compatibility.

The mapper uses an existing, well-tested presence guard so missing API fields are silently skipped. The metric is off by default on fresh installs and is placed correctly below the fold when enabled. Tests cover the mapper value, the disabled-by-default state, and the expanded-section ordering. Docs, README, and all five change sites are consistent with each other.

No files require special attention.

<sub>Reviews (1): Last reviewed commit: ["fix(claude): restore the weekly Sonnet l..."](https://github.com/robinebers/openusage/commit/3e07332538adf7b3dcc4581b26628a0a02fad4a8) | [Re-trigger Greptile](https://app.greptile.com/api/retrigger?id=40074622)</sub>

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ OpenUsage shows how much of your AI coding plans you've used: session and weekly
 ## Supported Providers
 
 - [**Antigravity**](docs/providers/antigravity.md) — Gemini Pro/Flash and Claude model quotas
-- [**Claude**](docs/providers/claude.md) — session, weekly, extra usage, local daily spend (ccusage)
+- [**Claude**](docs/providers/claude.md) — session, weekly, Sonnet, extra usage, local daily spend (ccusage)
 - [**Codex**](docs/providers/codex.md) — session, weekly, credits, local daily spend (ccusage)
 - [**Cursor**](docs/providers/cursor.md) — credits, total/auto/API usage, requests, on-demand, per-day spend
 - [**Devin**](docs/providers/devin.md) — weekly and daily quota, extra usage balance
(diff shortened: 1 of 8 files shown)


validatedev:
@codex review

Issue #753, opened 2026-06-26
Claude: delegate token refresh to the `claude` CLI when the stored refresh token is missing/revoked


robinebers:
## Context

Follow-up to #738 ("Claude - token expired"). The first fix (keychain-first credential selection + 429 backoff with last-good usage) restores the behavior of the legacy plugin and stops a stale `~/.claude/.credentials.json` from outranking the live keychain. But it leaves one robustness gap unaddressed.

OpenUsage refreshes the Claude OAuth token only via its own stored `refreshToken` (proactive near-expiry + reactive on 401, in `ClaudeProvider.fetchLiveUsage`). If the credential blob it reads has **no usable refresh token** (or a server-revoked one), every access-token expiry becomes an unrecoverable `tokenExpired` until an external `claude` run rewrites a fresh access token. Because Claude access tokens live ~1 hour, the menu bar then flaps: works while the user is active in Claude, fails ~10–30 min after they go idle.

## What the most-trusted monitors do

Surveying the field (ranked by stars + open-issue health):

- **[steipete/CodexBar](https://github.com/steipete/CodexBar) (15.4k★, no open issue on single-account token expiry)** — its primary refresh path is **delegated to the `claude` CLI**: `ClaudeOAuthDelegatedRefreshCoordinator` checks `isClaudeCLIAvailable`, "touches auth" so the CLI refreshes its own token (the CLI owns the refresh token + client credentials), then re-reads the keychain and verifies it changed. It wraps this in a cooldown (5 min default) and a `ClaudeOAuthRefreshFailureGate` (transient 5min→6h exponential vs terminal, keyed on a credential fingerprint so a re-login clears the block).
- **[griffinmartin/opencode-claude-auth](https://github.com/griffinmartin/opencode-claude-auth) (1.1k★)** and **[dotCipher/opencode-claude-bridge](https://github.com/dotCipher/opencode-claude-bridge)** — 3-layer refresh: re-read keychain → refresh stored token → **run the `claude` CLI** as the fallback.
- **[m13v/claude-meter](https://github.com/m13v/claude-meter)** — deliberately does no self-refresh and bails with "run `claude` once" (the honest version of OpenUsage's current bug).

## Proposed work

Add a CLI-delegated refresh layer for Claude, gated and cooldown-limited:

[...]

Issue #738, opened 2026-06-25
Claude - token expired


kristiankruse:
### Description

I've found some closed issues related to this problem about OU not seeing Claude logged in, but it's still not fixed in 0.6.28 or 0.7.0-beta.15.

I tried to follow some closed issues' suggestions around Keychain over credentials, but nothing fixed it, and the new beta didn't help. It worked fine until a week or so ago.

### Steps to Reproduce

1. Open OpenUsage
2. Claude not working
3. No matter how much I log in and out of Claude in terminal or desktop app, it's not working.

### Expected Behavior

After logging into Claude it should show usage.

### Actual Behavior

See screenshot.

### OpenUsage Version

0.7.0-beta.15

### macOS Version

26.6

### Mac Type

Apple Silicon

### Screenshots

<img width="666" height="428" alt="Image" src="https://github.com/user-attachments/assets/19d89bc4-3b41-4d37-a3ee-c15800285e90" />

robinebers:
Thanks for reporting this. Could you send over your log file? It will help us fix this.

`/Users/rebers/Library/Logs/OpenUsage/OpenUsage.log` ideally via email to `rob@sunstory.com`

kristiankruse:
Sent

kristiankruse:
Quick Claude bug report:
The log confirms both credential sources are reading tokens that have already been rotated by the Claude app. The rate-limited workaround (ccusage path) works because it actually invokes claude CLI fresh. A possible fix would be for OpenUsage to re-read credentials on tokenExpired rather than immediately failing both sources, or to always prefer the ccusage CLI path as primary for Claude.

robinebers:
> Quick Claude bug report: The log confirms both credential sources are reading tokens that have already been rotated by the Claude app. The rate-limited workaround (ccusage path) works because it actually invokes claude CLI fresh. A possible fix would be for OpenUsage to re-read credentials on tokenExpired rather than immediately failing both sources, or to always prefer the ccusage CLI path as primary for Claude.

Thank you! This will likely be fixed today, but there's still one sledge-hammer approach (the same you documented) which we might push too if the next beta remains unstable in #753.

robinebers:
@kristiankruse [v0.7.0](https://github.com/robinebers/openusage/releases/tag/v0.7.0) stable is now out with this fix included.

Let me know if it fixes it!

robinebers:
[...]

PR #773, opened 2026-06-27
feat(claude): delegate token refresh to the claude CLI when no usable refresh token (#753)

robinebers:
## TL;DR

When OpenUsage's stored Claude credential has an expired access token and **no usable refresh token**, it could never recover on its own and dead-ended on "token expired" until the app restarted. This PR adds a backstop: delegate the refresh to the `claude` CLI (which owns the refresh token + client creds), then re-read the keychain/file and verify the credential actually rotated — all gated by a cooldown and a transient-vs-terminal failure gate.

## What was happening

- OpenUsage refreshes Claude tokens in-process using the refresh token it can read. But the source it reads (keychain/file) sometimes carries an access token with **no usable refresh token** (the #738 shape).
- In that case an in-process refresh is impossible, so the provider threw `tokenExpired` and kept doing so on every refresh — it could only self-heal if the user manually re-ran `claude` *and* OpenUsage happened to re-read the new creds.
- There was no mechanism to ask the tool that *does* own the refresh token (the `claude` CLI) to refresh on our behalf.

## What this changes

- **`ClaudeCredentialFingerprint`** — a token-free, log-safe snapshot of the stored credential: `SHA256(accessToken + ":" + expiresAt)` plus the credentials file's mtime/size. Used to detect whether a refresh actually rotated the token. Never logs or stores a token value.
- **`ClaudeDelegatedRefreshCoordinator`** — resolves the `claude` binary (`CLAUDE_CLI_PATH` override, then `~/.claude/local/claude`, `~/.local/bin/claude`, Homebrew/usr-local, then a bare `claude` probed against the enriched PATH a GUI app needs). Runs a **non-interactive `claude --version`** touch, then polls the credential at 0.2/0.5/0.8s and returns `attemptedSucceeded` only when the fingerprint changed. `cliUnavailable` short-circuits *before* consuming the cooldown; a UserDefaults cooldown (20s reserve, 5 min on success) plus single-flight prevents stampedes.
- **`ClaudeRefreshFailureGate`** — distinguishes **terminal** failures (`invalid_grant`: no time expiry, monotonic — a later transient can't downgrade it) from **transient** ones (network/5xx/timeout: exponential back-off `5min·2^(n-1)` capped at 6h). Either block clears when the on-disk fingerprint changes (external re-login), rechecked at most once per 15s; transient also auto-unblocks by time. State persists to UserDefaults so a revoked token stays blocked across relaunches.
- **Wired into `ClaudeProvider`** — a terminal block with unchanged creds short-circuits the whole refresh with `tokenExpired` and **no network call**; the dead-end path and any auth-expiry from the live fetch attempt the delegated refresh and, on success, re-read candidates and retry the fetch once; transient transport errors back the gate off. Coordinator + gate are injectable via `init` (default-constructed around the auth store, like `ccusageRunner`) so tests substitute fakes.
- **Docs** — `docs/providers/claude.md` explains the delegated refresh and its cooldown/back-off in plain language, including the `CLAUDE_CLI_PATH` override.

## Tests

- `ClaudeRefreshFailureGateTests` — exponential back-off capped at 6h; transient expires by time; terminal never expires by time and a later transient doesn't downgrade it; changed fingerprint unblocks both kinds; unchanged stays blocked; 15s recheck throttle; `recordSuccess` clears; UserDefaults round-trip across instances. Suite-isolated `UserDefaults`.
- `ClaudeDelegatedRefreshCoordinatorTests` — `cliUnavailable` when nothing resolves (and does **not** consume cooldown); `skippedByCooldown` within the window then allowed after; `attemptedSucceeded` only when the post-touch fingerprint differs (→ 5-min cooldown); touch ran but unchanged → `attemptedFailed` (short cooldown); `CLAUDE_CLI_PATH` honored; single-flight shares one launch.
- `ClaudeProviderTests` (extended) — expired access token + no refresh token + coordinator succeeds (mock rewrites the file) → live usage fetched (the #753/#738 regression); coordinator fails / CLI unavailable → `tokenExpired`; terminal-blocked + unchanged creds → **no usage API call**. No token is logged on any path.
- `swift build` + full `swift test` green (511 tests, 1 skipped — the opt-in live-network test).

## Heads-up

- **Non-interactive touch chosen** per owner decision: the touch is `claude --version`, and correctness rides on the fingerprint-changed check rather than assuming the command rotates the token. If `--version` proves insufficient in the field (doesn't trigger the CLI's lazy refresh), the documented escalation is a PTY-driven `claude /status` — there's an explicit `// TODO(#753)` marking this. No PTY was built.
- Mirrors CodexBar's coordinator + gate, simplified for OpenUsage.
- Whether `claude --version` reliably triggers a refresh in practice is the open empirical question; the design fails safe either way (unchanged fingerprint → `attemptedFailed`, never a false success).

---

> [!NOTE]
> **Medium Risk**
> Touches authentication refresh, keychain/file reads, and subprocess execution of the Claude CLI; logic is complex but heavily tested and fails closed without assuming `--version` always rotates tokens.
> 
> **Overview**
> Adds a **delegated refresh** path for Claude when OpenUsage cannot refresh in-process (missing/revoked refresh token on the keychain/file it reads).
> 
> **New building blocks:** token-free `ClaudeCredentialFingerprint` (hash + optional file mtime/size when file is preferred); `ClaudeDelegatedRefreshCoordinator` runs non-interactive `claude --version`, polls fingerprints, single-flight + UserDefaults cooldowns, resolves CLI via `CLAUDE_CLI_PATH` / common paths / `which`; `ClaudeRefreshFailureGate` persists terminal vs transient blocks (exponential backoff, fingerprint-based unblock).
> 
> **`ClaudeProvider` wiring:** terminal-block short-circuit skips network unless CLI can still attempt; auth failures and no-refresh-token paths try CLI delegation before recording terminal failure; transient blocks skip hammering the usage API while still allowing credential recovery; after CLI success, re-read **preferred** candidate (not same stale source). `ClaudeAuthStore` exposes `currentFingerprint()` / `credentialsFilePath()`.
> 
> **Dev UX:** Settings Debug section (non-release updater builds) for full refresh, delegated CLI test, credential dump, clear gate; `WidgetDataStore.debugProvider`.
> 
> **Docs/tests:** `docs/providers/claude.md` user-facing explanation; extensive unit tests for gate, coordinator, and provider regressions.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5c0dd56cbb1e3894d26bda669b79991c01a89d31. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/OpenUsage/Providers/Claude/ClaudeAuthStore.swift
+++ b/Sources/OpenUsage/Providers/Claude/ClaudeAuthStore.swift
@@ -195,11 +195,19 @@ struct ClaudeAuthStore: Sendable {
     /// A token-free fingerprint of the credential currently in the preferred source (keychain, else
     /// file), used to detect whether an external `claude` re-login rotated the stored token. Re-reads
     /// the candidates fresh (nothing cached), so it reflects the live on-disk state.
+    ///
+    /// File metadata (mtime/size) is included ONLY when the file is the preferred source. When the
+    /// keychain is preferred, the fingerprint is just the keychain token hash — otherwise a touch that
+    /// only rewrites the file (e.g. `claude --version` updating the file's mtime without rotating the
+    /// keychain token OpenUsage actually reads) would change the fingerprint and yield a false
+    /// "rotated" signal, masking the still-expired keychain token and triggering a 5-min CLI cooldown
+    /// that blocks real recovery.
     func currentFingerprint() -> ClaudeCredentialFingerprint {
-        let oauth = orderedStoredCandidates().first?.oauth
+        let preferred = orderedStoredCandidates().first
+        let filePath = (preferred?.source == .file) ? credentialsPath() : nil
         return ClaudeCredentialFingerprint.make(
-            oauth: oauth,
-            credentialsPath: credentialsPath(),
+            oauth: preferred?.oauth,
+            credentialsPath: filePath,
             files: files
         )
     }
--- a/Sources/OpenUsage/Providers/Claude/ClaudeAuthStore.swift
+++ b/Sources/OpenUsage/Providers/Claude/ClaudeAuthStore.swift
@@ -186,6 +186,24 @@ struct ClaudeAuthStore: Sendable {
         envText("CLAUDE_CONFIG_DIR")
     }
 
+    /// The resolved `~/.claude/.credentials.json` path (honoring `CLAUDE_CONFIG_DIR`). Exposed so the
+    /// delegated-refresh layer can fingerprint the file's mtime/size; the keychain source has no path.
+    func credentialsFilePath() -> String {
+        credentialsPath()
+    }
+
+    /// A token-free fingerprint of the credential currently in the preferred source (keychain, else
+    /// file), used to detect whether an external `claude` re-login rotated the stored token. Re-reads
+    /// the candidates fresh (nothing cached), so it reflects the live on-disk state.
+    func currentFingerprint() -> ClaudeCredentialFingerprint {
+        let oauth = orderedStoredCandidates().first?.oauth
+        return ClaudeCredentialFingerprint.make(
+            oauth: oauth,
+            credentialsPath: credentialsPath(),
+            files: files
+        )
+    }
+
     // Resolved OAuth endpoint strings before URL validation. The suffix is derived from the same
     // env-var branching as the URLs but never depends on URL validity, so the (non-throwing) keychain
     // candidate path can read it without risking a throw.
--- a/Sources/OpenUsage/Providers/Claude/ClaudeCredentialFingerprint.swift
+++ b/Sources/OpenUsage/Providers/Claude/ClaudeCredentialFingerprint.swift
@@ -1,0 +1,70 @@
+import CryptoKit
+import Foundation
+
+/// A `Sendable` wrapper around `UserDefaults` for the delegated-refresh layer's persisted state. The
+/// `Foundation` class is thread-safe but not marked `Sendable`, so it can't be a stored property of a
+/// `Sendable` struct without this shim. The delegated-refresh coordinator genuinely crosses isolation
+/// (its touch runs in a detached `Task`), so the wrapper is the minimal escape hatch.
+struct SendableUserDefaults: @unchecked Sendable {
+    let defaults: UserDefaults
+    init(_ defaults: UserDefaults) { self.defaults = defaults }
+}
+
+/// A token-free, log-safe snapshot of "which credential is currently on disk", used to detect whether
+/// an external `claude` re-login actually rotated the stored credentials. We compare fingerprints
+/// before and after delegating a refresh to the `claude` CLI: an unchanged fingerprint means the CLI
+/// touch did NOT rotate the token (so the attempt failed), a changed one means it did (success).
+///
+/// NEVER carries a token value. `tokenHash` is a SHA256 of `accessToken + ":" + expiresAt`, so it
+/// changes when either the access token or its expiry rotates but can never be reversed to the token.
+/// The file `mtime`/`size` are a cheap secondary signal for the file source (the keychain has no file,
+/// so those stay nil for keychain-sourced credentials and the token hash carries the whole signal).
+struct ClaudeCredentialFingerprint: Codable, Equatable, Sendable {
+    var tokenHash: String?
+    var fileMTimeMs: Int?
+    var fileSize: Int?
+
+    /// Build a fingerprint from the OAuth blob OpenUsage already holds (so this never triggers an extra
+    /// Security.framework keychain prompt) plus optional file metadata for the file-backed source.
+    static func make(
+        oauth: ClaudeOAuth?,
+        credentialsPath: String?,
+        files: TextFileAccessing
+    ) -> ClaudeCredentialFingerprint {
+        ClaudeCredentialFingerprint(
+            tokenHash: tokenHash(for: oauth),
+            fileMTimeMs: nil,
+            fileSize: nil
+        ).withFileMetadata(path: credentialsPath)
+    }
+
+    /// SHA256 of `accessToken + ":" + (expiresAt ?? 0)`. Returns nil when there is no access token. The
+    /// value is a hex digest — never the token itself — so it is safe to persist and (in principle) log.
+    static func tokenHash(for oauth: ClaudeOAuth?) -> String? {
+        guard let accessToken = oauth?.accessToken, !accessToken.isEmpty else { return nil }
+        let expiresAt = oauth?.expiresAt ?? 0
+        // Format the expiry without scientific notation / locale drift so the hash is stable.
+        let material = "\(accessToken):\(Int64(expiresAt.rounded()))"
+        let digest = SHA256.hash(data: Data(material.utf8))
+        return digest.map { String(format: "%02x", $0) }.joined()
+    }
+
+    /// Returns a copy with `fileMTimeMs`/`fileSize` filled from the credentials file at `path` (if any).
+    /// Metadata is read via `FileManager` directly — only the real local accessor exposes a filesystem
+    /// path, so test doubles simply leave these nil and rely on the token hash, which is the real signal.
+    private func withFileMetadata(path: String?) -> ClaudeCredentialFingerprint {
+        guard let path, !path.isEmpty else { return self }
+        let expanded = (path as NSString).expandingTildeInPath
+        guard let attributes = try? FileManager.default.attributesOfItem(atPath: expanded) else {
+            return self
+        }
+        var copy = self
+        if let modified = attributes[.modificationDate] as? Date {
+            copy.fileMTimeMs = Int(modified.timeIntervalSince1970 * 1000)
+        }
+        if let size = attributes[.size] as? NSNumber {
+            copy.fileSize = size.intValue
+        }
+        return copy
+    }
+}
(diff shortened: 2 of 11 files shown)


robinebers:
Bugbot found a real bug: on the `sessionExpired`/`tokenExpired` catch path in `probe()`, `recordTerminalAuthFailure` was called before `delegatedRefreshIfPossible`, so the gate's 15s recheck throttle (set by `lastRecheckAt = now`) blocked the immediate `shouldAttempt` call and the CLI touch never ran. Any usable-looking-but-revoked refresh token (`invalid_grant` → `sessionExpired`) dead-ended even though the `claude` CLI could have rotated the credential.

Fixed in 4331259 by swapping the order — try `delegatedRefreshIfPossible` first; only record the terminal block if delegation also fails. Added a regression test (`testDelegatedRefreshRecoversWhenRefreshTokenRevokedAndCLISucceeds`) that exercises the revoked-refresh-token path with a CLI touch that rotates the credential. All 512 tests pass locally.

robinebers:
Bugbot's second pass found two more issues, both fixed in 6a88794:

1. **HTTP 5xx not transient** — the catch block's comment said "connection failed / 5xx" should back off the gate, but the code only recorded transient for `ClaudeUsageError.connectionFailed`. A 5xx (`requestFailed(503)`) skipped the gate. Fix: also record transient for `requestFailed(statusCode)` in `500..<600`. 4xx non-auth (e.g. 404) is NOT transient.

2. **False success on file metadata** — `currentFingerprint()` always mixed the preferred keychain token hash with `.credentials.json` mtime/size, so a touch that only rewrites the file (e.g. `claude --version` updating mtime without rotating the keychain token) changed the fingerprint → `attemptedSucceeded` → 5-min CLI cooldown, while the keychain token OpenUsage actually reads stayed expired. Fix: only include file metadata when the FILE is the preferred source — matching the PR author's original intent ("those stay nil for keychain-sourced credentials").

Added 4 regression tests. All 516 tests pass locally.

robinebers:
Bugbot's third pass found two more issues, both fixed in c8281c1:

1. **Terminal block stops CLI retry** (Medium, #bc1ed54a) — when auth failed after delegated refresh returned `cliUnavailable` or `skippedByCooldown`, the code still recorded a terminal gate with no time expiry. The pre-check then skipped all work until the fingerprint changed, so installing the CLI or waiting out the coordinator cooldown never triggered another delegated attempt. Fix: added `coordinator.canAttempt(now:)` (CLI resolves AND not in cooldown). Both the short-circuit and `delegatedRefreshIfPossible` now treat a terminal block as escapable when the coordinator could attempt — the CLI gets a chance to rotate the credential even while the terminal block is active. The coordinator's own cooldown still rate-limits the actual touch.

2. **Terminal gate blocks good credentials** (High, #8898f938) — after a delegated CLI refresh updated the preferred keychain while a fallback file source was probed, recovery kept the stale file candidate (same-source preference), auth failed again, and a terminal gate was saved using the preferred source's (now valid) fingerprint. Later refreshes short-circuited without trying the good keychain. Fix: use the PREFERRED candidate (first) after recovery — the CLI typically rotates the preferred source, so the same-source candidate may still be stale.

Added 3 regression tests. All 519 tests pass locally.

robinebers:
Bugbot's fourth pass found two more issues, both fixed in d31f704:

1. **Retry fetch skips gate updates** (Medium, #88d01254) — after a successful delegated CLI recovery, the second `fetchLiveUsage` call sat inside the auth catch handler, so errors from the retry were not handled by the surrounding do/catch. A 5xx on the retry never invoked `recordTransientFailure`, and an auth failure after rotation skipped the terminal-recording path. Fix: wrap the retry in its own do/catch that routes through the same gate-update logic (extracted into `recordTransientForUsageError` helper). No second delegated refresh (avoids loops).

2. **Terminal skip ignores fresh creds** (Medium, #f6dcff9f) — the terminal short-circuit treated `!shouldAttempt` as "credentials unchanged," but `shouldAttempt` can be false for up to 15s solely because of the recheck throttle — even after an external re-login changed the fingerprint. With the CLI unavailable, refresh returned `tokenExpired` for up to 15s after the user re-logged in. Fix: added `forceRecheck` parameter to `shouldAttempt`; the short-circuit uses `forceRecheck: true` so an external re-login is detected immediately. The throttled path is still used by `delegatedRefreshIfPossible` for the hot refresh path.

Added 2 regression tests. All 521 tests pass locally.

robinebers:
Bugbot's fifth pass found one Low-severity issue, fixed in 6d560f7:

**Success cooldown ignores injected time** (Low, #ed400fe6) — `runTouchAndVerify` stamped the success cooldown with `now()` (the closure) while the short cooldown at attempt start used the `moment` passed via `attempt(now:)`. Callers that inject time only via the parameter got mismatched cooldown timestamps. Fix: stamp the success cooldown with `moment` for consistency. In production `moment == now()` so the ~1.5s of verify-poll delay is negligible vs the 5-min cooldown; in tests the timestamps are now consistent.

Added 1 regression test. All 522 tests pass locally.

robinebers:
Bugbot's sixth pass found one Medium issue, fixed in 325a689:

**Transient gate skips CLI recovery** (Medium, #737e65d5) — `delegatedRefreshIfPossible` was gated solely on `failureGate.shouldAttempt` or a terminal-block escape hatch. While a transient block was active (e.g. after a recent 503 on the usage API), auth failures skipped the CLI entirely and fell through to recording a terminal failure — even though the CLI is the intended recovery path when in-process refresh is impossible. The transient block (from a 5xx/connection failure on the usage API) is orthogonal to auth failures: it backs off the usage API, not the CLI. Fix: extend the escape hatch to ANY block type — `delegatedRefreshIfPossible` now tries the CLI whenever the coordinator can attempt, regardless of whether the gate's block is terminal or transient. The terminal-block escape hatch (#bc1ed54a) is subsumed by this. `shouldAttempt` is still called for its fingerprint-recheck side effect.

Added 1 regression test. All 523 tests pass locally.

robinebers:
Bugbot's seventh pass found two more issues, both fixed in 17f180f:

1. **Terminal check blocks main actor** (Medium, #1ba3ab5b) — while a terminal block was active, each `refresh()` called `failureGate.shouldAttempt(now:forceRecheck: true)` on the main actor, which synchronously ran `currentFingerprint()` (blocking `/usr/bin/security` keychain reads) even though credentials were just loaded off-main at the start of `refresh()`. Fix: moved the entire short-circuit check (`currentBlockStatus` + `shouldAttempt` + `canAttempt`) off-main via `loadOffMainActor`. The keychain read and filesystem checks now run on a background executor; the main actor just awaits the boolean decision.

2. **Transient gate skips API backoff** (Medium, #bc71842b) — the gate records exponential backoff on 5xx/connection errors, but `refresh()` only short-circuited for terminal blocks, so an active transient block still hit `/api/oauth/usage` on every periodic refresh. Fix: `fetchLiveUsage` skips the usage API call when a transient block is active, serving the last-good usage instead. The check is placed AFTER the delegated/in-process refresh steps so credential recovery still runs — only the usage API call is skipped.

Added 1 regression test. All 524 tests pass locally.

robinebers:
Bugbot's eighth pass found one Medium issue, fixed in 7915fd5:

**CLI probe runs before baseline** (Medium, #d30aea11) — when the `claude` binary was resolved only via PATH, `resolveCLI()` ran a non-interactive `claude --version` probe BEFORE `runTouchAndVerify` captured its baseline fingerprint. If that probe refreshed credentials, the baseline reflected the new state, the second touch looked unchanged, and the coordinator returned `attemptedFailed` even though rotation already happened. Fix: `commandExists` now uses `/usr/bin/which` to check PATH presence WITHOUT running `claude` — the actual `--version` touch (and its fingerprint verification) happens in `runTouchAndVerify` against a pre-touch baseline. Test runners updated to handle `/usr/bin/which` appropriately.

Added 1 regression test. All 525 tests pass locally.

robinebers:
Closing this — the delegated refresh touch (`claude --version`) doesn't actually trigger a token refresh, so the recovery path this PR adds is a no-op in practice.

Confirmed three ways:

1. **Debug button test** (this PR's own debug section): `claude --version` runs, exits 0, but the credential fingerprint is unchanged — no rotation. See the [debug log trace](https://github.com/robinebers/openusage/pull/773#issuecomment-4825136847).
2. **CodexBar #1287** (their maintainer's own finding): [`claude --version` doesn't ping the API so doesn't trigger a refresh](https://github.com/steipete/CodexBar/issues/1287), and their delegated refresh "always times out" — their touch is a separate `ClaudeStatusProbe`, not `--version`, and even that doesn't work reliably.
3. **claude-code #50412**: the CLI's refresh gate only fires within ~5 min of expiry or on a 401; `--version` does neither.

The PR's fail-safe design (fingerprint-verify, no false success) is correct, and the coordinator/gate infrastructure is sound. But the touch command doesn't rotate the credential, and the documented escalation (PTY-driven `claude /status`) is also unproven — CodexBar's similar approach times out per #1287. The direct `refresh_token` exchange alternative has desync risks (CodexBar #1161 — Anthropic rotates refresh tokens on every use, which breaks the CLI).

Reopen when either Anthropic fixes the CLI's refresh behavior (claude-code #31095, #50412) or a touch command is confirmed to reliably trigger it. The machinery is ready if a working touch materializes.

Refs: steipete/CodexBar#1287, steipete/CodexBar#1161, anthropics/claude-code#50412, anthropics/claude-code#31095.

PR #861, opened 2026-07-06
feat(kiro): add Kiro provider for monthly credit usage

Rich627:
## TL;DR
Adds Kiro as a new provider, tracking monthly credit usage and overage charges by reading the same local credentials `kiro-cli` already stores and calling the CodeWhisperer usage endpoint it uses internally.

## What was happening
OpenUsage didn't track Kiro CLI usage at all — no widget, no auth reading, no docs.

## What this changes
- New `Kiro/` provider module: `KiroAuthStore` (reads the access token + profile ARN from kiro-cli's local SQLite database), `KiroUsageClient` (calls AWS CodeWhisperer's `GetUsageLimits`), `KiroUsageMapper` (normalizes the response into a Credits progress row + an Overage Charges row when non-zero), and `KiroProvider` tying it together.
- Registers Kiro in `AppContainer`, `DefaultLayout` (enabled by default, alphabetically slotted between Grok and OpenRouter), and the provider links test fixture.
- Adds a Kiro SVG mark and `docs/providers/kiro.md`.
- Adds `Tests/OpenUsageTests/KiroProviderTests.swift` covering the mapper, auth store, and provider runtime.

## Heads-up
While reviewing, found and fixed two issues before opening this PR:
- The credits progress line was clamping `used` to `limit` (`min(used, limit)`), which would hide real overage counts — a user over their plan's limit would always see e.g. "50 of 50" instead of the true "55 of 50," even though the app's meter UI is built to show over-limit state. Fixed to pass the true usage through, and added a regression test.
- Removed `KiroAuthStore.isTokenExpired(_:)` — dead code, never called from `refresh()` or anywhere else.
- Minor: restored the missing trailing newline in `ErrorCategory.swift`.
- Screenshot pass caught a UI resolution miss where the Kiro descriptor looked for `Credits used` while the mapper emitted `Credits`; fixed so the dashboard renders live Kiro credit data instead of `No data`.

## Tests
- `swift build` — clean.
- `swift test --filter Kiro` — all 16 pass, including the new overage-clamping regression test.
- `swift test` (full suite) — same 4 pre-existing `Codex` failures as on `main` before this change (verified via stash), nothing newly broken.

## Screenshots
Before: Not applicable - Kiro is a new provider widget.

After:

Kiro provider dashboard widget


Rich627:
Added the Screenshots section to the PR description

robinebers:
I am closing this until there is proven demand

PR #865, opened 2026-07-07
fix(claude): prefer profile-scoped login over inference-only env token for live usage

### LINKED ISSUE #901: Claude live meters show "No data" when an inference-only CLAUDE_CODE_OAUTH_TOKEN shadows a real login
## Summary

Claude's live meters (Session / Weekly / Sonnet / Fable / Extra Usage) silently show **No data** for a signed-in Max user whenever an inference-only `CLAUDE_CODE_OAUTH_TOKEN` is present in the environment — even though a valid, profile-scoped keychain login exists and would return full usage.

## Steps to reproduce

1. Sign in normally with Claude Code (a keychain login with the `user:profile` scope, e.g. a Max plan).
2. Export a long-lived `CLAUDE_CODE_OAUTH_TOKEN` from `claude setup-token` in your shell profile (common for headless/CI use; it lingers in the interactive shell environment).
3. Launch OpenUsage.

**Expected:** Session / Weekly / etc. populate from the keychain login.
**Actual:** They all read **No data**, with no warning. Only the log-derived spend tiles (Today / Yesterday / Usage Trend) render.

## Root cause

- On macOS, `ProcessEnvironmentReader` falls back to the captured **login-shell** environment, so an ambiently-exported `CLAUDE_CODE_OAUTH_TOKEN` is visible to OpenUsage.
- That token is **inference-only**: it can run the model but returns **HTTP 403** on `GET /api/oauth/usage`.
- `ClaudeAuthStore.loadCredentialCandidates()` treated any `CLAUDE_CODE_OAUTH_TOKEN` as an override and returned a **single** candidate with `inferenceOnly = true`.
- `ClaudeProvider.probe()` then took the `.inferenceOnlyToken` branch and **skipped the usage call entirely** (zero requests to `api.anthropic.com`), so the meters went blank while the real keychain login was never consulted.

Diagnosed live: with the env token set, the app made no usage request at all; the keychain login (Max, `user:profile`, unexpired) returned full usage when queried directly.

## Proposed fix

Prefer any stored login that can actually read usage (keychain-first, then file) for the live usage call, keeping the inference-only env token only as a trailing fallback; when there is no live-capable stored login (a genuinely headless setup), behavior is unchanged. Fix proposed in #865.

joshuavial:
Fixes #901

## TL;DR

An inference-only `CLAUDE_CODE_OAUTH_TOKEN` in the environment shadowed a real keychain login, so the Claude Session/Weekly/Sonnet/Fable/Extra Usage meters silently showed **No data** even for a signed-in Max user. This makes OpenUsage prefer the profile-scoped login for the live usage call, keeping the env token only as a fallback.

## What was happening

- On macOS, OpenUsage reads Claude Code's credentials, and `ProcessEnvironmentReader` falls back to the captured **login-shell** environment so it can pick up exported keys.
- Many users have a long-lived `CLAUDE_CODE_OAUTH_TOKEN` (from `claude setup-token`) exported in their shell profile. It is **inference-only** — it can run the model but returns **HTTP 403** on `GET /api/oauth/usage`.
- `ClaudeAuthStore.loadCredentialCandidates()` treated any `CLAUDE_CODE_OAUTH_TOKEN` as an override and returned a **single** candidate with `inferenceOnly = true`.
- `ClaudeProvider.probe()` then took the `.inferenceOnlyToken` branch and **skipped the usage call entirely** — no request to `api.anthropic.com` at all — so Session/Weekly/Sonnet/Fable/Extra Usage rendered **No data**, with no warning. Only the log-derived spend tiles (Today/Yesterday/Usage Trend) kept working, which is exactly the symptom users saw: blank meters but a populated Usage Trend.
- The user's real keychain login (Max, `user:profile` scope, unexpired) returned full usage the whole time — it was just never consulted.

## What this changes

- `loadCredentialCandidates()`: when `CLAUDE_CODE_OAUTH_TOKEN` is set, it no longer returns only the env token. It now returns any stored login that can actually read usage (`canFetchLiveUsage`, keychain-first then file) **ahead** of the env token, with the inference-only env token kept as a trailing fallback for the refresh loop.
- If there is **no** live-capable stored login (a genuinely headless setup), the env token remains the sole candidate — behavior is unchanged, and the local-log spend tiles still load.
- The env-token fallback candidate is now sourced honestly as `.environment` (instead of borrowing the stored login's source), so the refresh-start diagnostics name the real source, and `save()` correctly no-ops instead of ever writing an env token back into the keychain under a borrowed source.
- Nothing is silenced: the env path stays explicit; only the credential **selected** for the usage fetch changes. Net effect is a strict improvement — the meters now populate whenever a usable login exists, and the headless case is untouched.

## Heads-up

- Two related edge cases remain where an ambient env token still suppresses the "Re-login for live usage" nudge (a scope-deficient real login, and the auth-expiry-to-env-fallback path). These are not regressions (behavior is no worse than before this change) and are a warning-policy decision rather than a code bug — worth a follow-up if you want the nudge surfaced on those paths.

## Tests

- Two regression tests added: `testEnvTokenDoesNotShadowProfileScopedStoredLogin` (env token + profile-scoped keychain login → the login is preferred, env token trails) and `testEnvTokenIsSoleCandidateWhenStoredLoginCannotReadUsage` (no live-capable login → env token is the sole inference-only candidate, unchanged).
- `swift build` clean; full `swift test` green (897 tests) under `TZ=UTC`.

## Screenshots

**Before** (installed build with the env token set)

image

**After** (this branch, same machine, `CLAUDE_CODE_OAUTH_TOKEN` still exported):

image


robinebers:
Correctly flagged but this is a great bug. Please create an issue for it and we'll tackle it asap.

robinebers (commented):
Deep-review pass. The diagnosis and fix are sound: I verified against the auth-store code that the old path really did return the env token as the *sole* candidate (so `probe()` skipped the usage call entirely — blank meters with a live keychain login sitting unused), and this change is a strict superset of that behavior. The env candidate keeps `inferenceOnly: true` so it can never trigger a usage call, sourcing it as `.environment` fixes both the diagnostics and the `save()` no-op, the headless case is provably unchanged, the regression tests pin both directions, and the docs update matches AGENTS.md's requirement. One must-fix, though — and it's invisible to GitHub's conflict detection:

**Must fix before merge: this no longer compiles against `main`.** #868 (merged after this PR was opened) removed `ClaudeAuthStore.canFetchLiveUsage` — it was a one-line wrapper over `liveUsageAvailability(_:) == .available`, deleted as dead code because nothing in production called it. This PR now calls it in production (`stored.filter(canFetchLiveUsage)`) and in both new tests. I test-merged this branch onto current `main`: the merge is **textually clean**, so nothing will flag it until CI fails to compile. After rebasing, port to the surviving seam:

- `let liveCapable = stored.filter { liveUsageAvailability($0) == .available }`
- In the test, replace `XCTAssertTrue(store.canFetchLiveUsage(candidates[0]))` with `XCTAssertEqual(store.liveUsageAvailability(candidates[0]), .available)`

Semantics are identical (including the deliberate "absent/empty scopes → available" permissiveness, which correctly lets older scope-less stored logins be preferred over the env token).

**Nit:** the env fallback borrows its display metadata from `stored.first` — if the first stored candidate is *not* live-capable but a later one is (e.g. an inference-only keychain login plus a profile-scoped file login), the fallback inherits `fullData`/plan metadata from the login you just decided not to prefer. `let base = liveCapable.first ?? stored.first` (computed after the filter) would borrow from the credential actually being used. Cosmetic — availability is decided by `inferenceOnly: true` either way, and the oauth-borrowing itself is carried over from the old code, not new.

**Process gates** (owner's call, just flagging): the PR carries `needs-approved-issue` and `needs-screenshots` — the Screenshots section describes the before/after in prose but has no images, and this change does alter what the meters display.

The Heads-up section's honesty about the two remaining nudge-suppression edge cases is appreciated — agreed they're pre-existing policy questions, not regressions from this change.

---

joshuavial:
Thanks for the thorough review — all addressed:

- **Must-fix (compile against `main`):** rebased onto current `main`; ported to the surviving seam exactly as suggested — `stored.filter { liveUsageAvailability($0) == .available }` in the store, and `XCTAssertEqual(store.liveUsageAvailability(candidates[0]), .available)` in the test. Builds clean; full suite green.
- **Nit:** implemented — `let base = liveCapable.first ?? stored.first` (computed after the filter), so the env fallback borrows plan metadata from the credential actually preferred. Added a regression test (`testEnvFallbackBorrowsMetadataFromThePreferredLiveCapableLogin`) covering the inference-only-keychain + profile-scoped-file case.
- **Issue:** filed #901; PR now linked via `Fixes #901`.
- **Screenshots:** added an after shot (Session/Weekly populated with the env token still set); before was the all-"No data" state.

robinebers:
@cursor please review the current PR head and fix any issues you find directly on this branch. Keep the change focused on #901, preserve the credential-source fallback behavior, and add or update regression tests for any fix.

robinebers:
LGTM

PR #868, opened 2026-07-07
refactor: DRY/dead-code/KISS cleanup from code-quality audit

robinebers:
**TL;DR** — A code-quality audit surfaced ~44 DRY / dead-code / KISS / SOLID findings; this resolves them in one net **−947 LOC** cleanup. No behavior change intended. Foundation for four follow-up god-object-split PRs (stacked on this branch).

## What was happening
- The same day-bucketing + per-day aggregation was copy-pasted across the Claude/Codex/Grok scanners, Cursor, and SpendTileMapper — the exact drift class behind the shipped ccusage false-zero bug.
- Claude & Codex duplicated ~90 lines of incremental-scan scaffolding; both files were over the repo's ~500 LOC rule.
- OpenRouter/ZAI auth stores were line-for-line copies; `ModelHoverState`/`TrendHoverState` were identical hover state machines.
- WidgetData decided tray/session behavior by matching a tile's **title string** and a hardcoded widget-ID list (renaming a tile silently broke the tray).
- Devin ran blocking credential file I/O on the MainActor (could stutter the popover).
- A pile of production-dead, test-only APIs.

## What this changes
- **Shared infra:** `DailyUsageAccumulator` + one `dayKey`; generic `IncrementalJSONLScanner<Item>` actor (both scanners now under 500 LOC); `UserAPIKeyStore`; merged `HoverPopoverState`; `ProviderParse.bool`/`unwrapGoKeyring`, `UserDefaults.bool(forKey:default:)`, `SystemClients.writePassword`, `SettingsScreen.inlineNotice`.
- **Provider mappers:** CodexUsageMapper `windowLine` (6 sites → 1); Cursor `CursorPlanUsageFacts` (predicates re-derived 3× → decoded once); ZAI period plumb-through.
- **Views:** `TransientPill`, `DismissableHintCard`, `ShareCardChrome` + `renderAndCopy`, `VisibilityGatedTimeline`, shared neighbor-aware condensing, single-provider `APIKeysSection`.
- **SOLID/correctness:** Devin credential I/O off the MainActor; WidgetData `traySuffix`/`isSessionWindow` descriptor opt-ins replace the title match + ID list.
- **Dead code removed:** `source`/`Source` enum ×4, `maskedKeyPreview`, `Pace.status`, `menuBarPrimaryText`, `currentModelIsFallback`, `maxMode`, `isMetricExpanded`, several unused props — tests ported to the live seams.

## Heads-up
[...]

PR #869, opened 2026-07-07
refactor(LayoutStore): extract PopoverNavigationStore + generic TransientNotice

robinebers:
**TL;DR** — First of four god-object-split follow-ups. Pulls in-popover navigation and the three transient-notice machines out of `LayoutStore` into focused sub-stores, **preserving its public API** so no call sites change. Stacked on #868.

> Base is the cleanup branch (#868), so this PR's diff shows only the split. GitHub CI ("Build and Test") runs on base=main PRs; it'll run here once #868 merges and this retargets to main. Verified locally in the meantime (build + 883 tests + app launch). Cursor Bugbot runs regardless.

## What was happening
`LayoutStore` (1200+ LOC) mixed layout state with (a) popover screen routing + slide bookkeeping and (b) three copy-pasted transient-pill state machines (pin-denial, "Copied to clipboard", Customize notice), each a value + replay-trigger + clear `Task`.

## What this changes
- **`PopoverNavigationStore`** — owns `screen` (+ its slide-bookkeeping `didSet`), `isEditing`, `customizeProviderID`, `screenSlideFrom/ID`. `LayoutStore` forwards its existing surface to it, so the ~59 call sites across 8 files are unchanged.
- **`TransientNotice<Value>`** — one reusable auto-clearing pill; the three machines collapse to three instances. Public `pinLimitNotice` / `shareConfirmation` / `customizationNotice`(+`Tone`) / `*Trigger` forward to these. The clear `Task`s are now encapsulated in the notice type.

## Heads-up
- API-preserving by design: navigation state is exposed through `LayoutStore`'s existing properties (forwarding to the sub-store) rather than a renamed `layout.navigation.screen` path — a 59-site rename of SwiftUI observation wasn't worth the regression risk for an organizational change. The state genuinely lives in the sub-stores.
- SwiftUI observation is preserved via the standard nested-`@Observable` access-tracking pattern; validated by the `PopoverScreen`/navigation forwarding tests + an app run (screen state, slide counter, Customize L2 reset all exercised).

## Tests
- `swift test` — 883 tests, 0 failures. Release build launches; providers refresh cleanly; no crashes.

---

> [!NOTE]
> **Low Risk**
> API-preserving refactor of UI-only navigation and transient notices; no persistence or layout mutation logic changes, with new unit tests for timer behavior.
> 
> **Overview**
> **`LayoutStore` shrinks** by moving two concerns into focused types while **call sites stay on the same properties** (`screen`, `isEditing`, pin/share/customize notices, triggers).
> 
> **`PopoverNavigationStore`** now owns `PopoverScreen`, screen-switch slide bookkeeping (`screenSlideFrom` / `screenSlideID`), Customize L2 (`customizeProviderID`), and `isEditing`. `LayoutStore` holds a private `navigation` instance and forwards getters/setters so SwiftUI observation and ~59 existing references do not need renames.
> 
> **`TransientNotice<Value>`** replaces three identical patterns (value + replay `trigger` + auto-clear `Task`) with one generic helper. Pin denial, share confirmation, and Customize pills use separate instances; `present` / `clear` replace inline timer code. Customize message and **tone** are bundled in **`CustomizationNoticeContent`** so they clear atomically.
> 
> **`TransientNoticeTests`** add async coverage for auto-clear, re-present timer restart, and immediate `clear()`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7f1af892eab1f001ea72e82dec9825cc558e8b89. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/OpenUsage/Stores/LayoutStore.swift
+++ b/Sources/OpenUsage/Stores/LayoutStore.swift
@@ -10,8 +10,9 @@ final class LayoutStore {
 
     /// In-popover navigation (screen, Customize master/detail, screen-switch slide). Its own store so
     /// screen routing isn't tangled with layout state; the `screen`/`isEditing`/`customizeProviderID`/
-    /// `screenSlide*` surface below forwards to it, so existing call sites are unchanged.
-    let navigation = PopoverNavigationStore()
+    /// `screenSlide*` surface below forwards to it, so existing call sites are unchanged. Private so the
+    /// forwarding surface stays the ONLY spelling — two live paths to the same state invites drift.
+    private let navigation = PopoverNavigationStore()
 
     /// Which in-popover screen is showing. Drives the footer buttons, the Esc handler, and the
     /// popover-closed reset alike.
@@ -57,15 +58,12 @@ final class LayoutStore {
     /// stay open across popover closes and app restarts.
     private(set) var expandedProviderIDs: Set<String>
 
-    /// Transient explanation for a denied pin attempt (the WhatsApp-style "you can only pin N chats"
-    /// feedback). Set by `notePinDenied`, cleared automatically a few seconds later; the popover footer
-    /// renders it in place of the pin counter. Never persisted.
     /// The three transient popover pills, each an auto-clearing `TransientNotice` (was three copy-pasted
     /// value+trigger+clearTask machines). The public `pinLimitNotice`/`shareConfirmation`/
     /// `customizationNotice` surface below forwards to these, so call sites are unchanged.
     private let pinNotice = TransientNotice<String?>(clearedValue: nil, timeout: .seconds(3))
     private let shareNotice = TransientNotice<Bool>(clearedValue: false, timeout: .seconds(2.5))
-    private let customizationNoticeStore = TransientNotice<CustomizationNoticeContent?>(clearedValue: nil, timeout: .seconds(2.5))
+    private let customizeNotice = TransientNotice<CustomizationNoticeContent?>(clearedValue: nil, timeout: .seconds(2.5))
 
     /// Transient explanation for a denied pin attempt; the popover footer renders it in place of the pin
     /// counter. Set by `notePinDenied`, auto-cleared a few seconds later.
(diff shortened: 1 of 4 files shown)


robinebers (commented):
Deep-review pass (DRY / SOLID / KISS focus), verified against the actual diff vs #868's branch. The core claims hold up: the nested-`@Observable` forwarding is correctly applied (every forwarder reads a tracked stored property on the sub-store, so SwiftUI invalidation and the `withObservationTracking` blocks in `StatusItemController` keep working), the `screen` `didSet` slide bookkeeping moved verbatim with the same ordering and the Customize L2 reset intact, and all three `TransientNotice` instances preserve their durations (3s / 2.5s / 2.5s), trigger bumps, and cancel-prior-task semantics. No blockers.

**Should fix before merge:**

1. **Orphaned doc comment in `LayoutStore.swift` (~line 60).** The old pin-denial doc block ("Transient explanation for a denied pin attempt…") was left dangling above the new "three transient popover pills" comment when its property moved, and it duplicates the real `pinLimitNotice` doc a few lines below. Dead text from the diff — delete it.
2. **`navigation` is internal alongside the full forwarding facade** (`LayoutStore.swift:14`). That ships two parallel APIs to the same state (`layout.screen` and `layout.navigation.screen`), which invites exactly the call-site divergence the facade exists to prevent. Nothing outside `LayoutStore` uses `.navigation` today, and the notice sub-stores got this right (`private let`). Make it `private let navigation`.

**Nits:**

- Naming inconsistency: `pinNotice`, `shareNotice`, but `customizationNoticeStore` — pick one convention.
- One real (if unobservable) state delta: `customizationNoticeTone` is now `value?.tone ?? .positive`, so it snaps to `.positive` the instant the notice clears, where the old code kept the last tone sticky. Tone is only read while the notice is non-nil, so nothing is user-visible — but it's the single behavioral difference in the refactor and deserves a one-line comment.
- `PopoverNavigationStore.swift` imports SwiftUI but only needs `Observation`; `TransientNotice.swift`'s Foundation import is likewise unused.
- `@ObservationIgnored` on the `let clearedValue` / `let timeout` in `TransientNotice` is redundant (`let`s are never tracked); only `clearTask` needs it.
- The new types have no direct tests — coverage is entirely through the LayoutStore facade. That indirection is actually good evidence of API preservation, but a small `TransientNotice` test (re-present cancels the pending clear; `clear()` cancels the timer) would lock in the timer semantics all three pills now share.

Fixes for the above incoming on this branch.

---

PR #870, opened 2026-07-07
refactor(WidgetDataStore): extract QuotaNotificationEvaluator

robinebers:
**TL;DR** — Second god-object split. Moves the quota pace-notification subsystem out of `WidgetDataStore` into a focused `QuotaNotificationEvaluator`; the store drops 634 → **502 LOC**. Internal-only (public API unchanged). Stacked on #869.

> Stacked on the LayoutStore split (#869 → #868). GitHub CI runs once the stack retargets to main; verified locally (build + 883 tests). Bugbot runs regardless.

## What was happening
`WidgetDataStore` bundled three concerns: refresh orchestration, a complete quota-notification subsystem (`evaluateNotifications` + `post` + `NotificationState` dedup + ~70 lines of trace-formatting helpers), and the `MetricLine`→`WidgetData` resolver.

## What this changes
- **`QuotaNotificationEvaluator`** owns the dedup state, the fire/deliver/commit decision, the notification build (title/subtitle/body), and the trace helpers.
- `WidgetDataStore.evaluateNotifications(now:)` stays (same signature) but now just gathers the pass's enabled/bounded/visible metrics and delegates; delivery (`postNotification`) and provider display name pass in as `@MainActor` closures, so the evaluator doesn't depend on the store's providers.
- State pruning is preserved: the evaluator rebuilds state only from the metrics it's given, so a disabled/removed metric starts fresh.

## Heads-up
- Behavior-preserving; the extracted logic is byte-for-byte the same decision path, just relocated.
- The `MetricLine` resolver (`resolve`/`resolveText`/`textData`) was left in place — the notification extraction alone brings the file to the ~500 rule, and the resolver is more entangled with the store's snapshot state (a separate change if desired).

## Tests
- `swift test` — 883 tests, 0 failures. `WidgetDataStoreNotificationTests` drive `evaluateNotifications` end-to-end (fire, dedup, prime, reset-jitter, prune) through the new evaluator.

---

> [!NOTE]
> **Low Risk**
> Internal refactor with the same public API; the only noted behavior shift is frozen metric snapshots within a single notification pass, which is unlikely to matter in normal periodic evaluation.
> 
> **Overview**
> Moves quota **pace-notification** logic out of `WidgetDataStore` into a new **`QuotaNotificationEvaluator`** that holds per-metric dedup state, `PaceNotificationLogic` transitions, delivery/commit after successful posts, and debug trace helpers. **`evaluateNotifications(now:)`** is unchanged publicly: the store still filters enabled layout metrics to bounded meters, then delegates with injected **`postNotification`** and provider display-name closures.
> 
> **Intentional pass behavior change:** metric `WidgetData` is snapshotted for the whole pass before any delivery `await`, instead of re-calling `data(for:)` between notifications—so a mid-pass refresh cannot alter later metrics in the same evaluation pass.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 43241b411486bf29285429764febd27e63edade1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/OpenUsage/Stores/QuotaNotificationEvaluator.swift
+++ b/Sources/OpenUsage/Stores/QuotaNotificationEvaluator.swift
@@ -1,0 +1,169 @@
+import Foundation
+
+/// Owns the quota pace-notification subsystem: the per-metric dedup state, the fire/deliver/commit
+/// decision, and the debug trace. Split out of `WidgetDataStore` (which orchestrates refresh + resolves
+/// `MetricLine`s) so the notification concern is self-contained. The store gathers each pass's enabled,
+/// bounded, visible metrics and calls `evaluate`; delivery + the provider display name come in as
+/// closures so this type stays independent of the store's providers.
+///
+/// Deduped per metric per reset window; the no-trustworthy-pace states (no data, fresh session, level
+/// bands) never fire. State for metrics not passed this pass is pruned, so re-enabling/re-adding a
+/// metric starts fresh rather than carrying a stale "already fired" flag.
+@MainActor
+final class QuotaNotificationEvaluator {
+    /// One enabled, bounded, visible metric for this pass, already resolved by the store.
+    struct Metric {
+        let key: String
+        let providerID: String
+        let data: WidgetData
+    }
+
+    private var notificationState: [String: NotificationState] = [:]
+
+    /// Evaluate every metric for a quota pace milestone and deliver a notification for any that just
+    /// crossed one, via `post` (id-prefix, title, subtitle, body → delivered?). `providerName` maps a
+    /// provider id to its display name for the subtitle.
+    func evaluate(
+        metrics: [Metric],
+        toggles: PaceNotificationToggles,
+        now: Date,
+        providerName: @MainActor (String) -> String,
+        post: @MainActor (String, String, String, String) async -> Bool
+    ) async {
+        var nextState: [String: NotificationState] = [:]
+        for metric in metrics {
+            let key = metric.key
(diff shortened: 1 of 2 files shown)


robinebers (commented):
Deep-review pass (DRY / SOLID / KISS focus). I verified the "byte-for-byte decision path" claim mechanically — stripped indentation and diffed the moved blocks between old and new files. The fire → deliver → commit ordering, dedup key, failed-delivery revert, reset-jitter gating, pruning-by-omission, notification copy, and all ten trace helpers diff clean. The evaluator is correctly created once (`@ObservationIgnored private let`), so dedup state persists across passes, and the closure injection captures `[providersByID]` by value with no retain of the store. Tests still drive the real store → evaluator path. **No blockers, no should-fixes.** Two nits:

1. **Stale doc reference** — `WidgetDataStore.swift:161` still says "Deduped per metric per reset window via `notificationState`", but `notificationState` now lives privately inside the evaluator. Point it at the evaluator's dedup state.
2. **Intra-pass snapshot semantics quietly changed** (informational — arguably an improvement, but it contradicts the unqualified "byte-for-byte" claim): the new code gathers all enabled/bounded metrics eagerly before the first `await`, where the old loop evaluated `data(for:)` lazily between delivery suspension points. If a refresh lands or a provider is disabled mid-pass, old code let later metrics see the update; new code decides from a pass-start snapshot. More consistent, unobservable by tests — but worth a one-line comment so the next reader knows it's deliberate.

---

PR #871, opened 2026-07-07
refactor(StatusItemController): extract StatusItemImageUpdater

robinebers:
**TL;DR** — Third god-object split. Pulls the menu-bar strip render loop out of `StatusItemController` into a focused `StatusItemImageUpdater`. Behavior-preserving (logic relocated verbatim). Stacked on #870.

> Stacked (#870 → #869 → #868). GitHub CI runs once the stack retargets to main; verified locally (build + 883 tests + **app launch**). Bugbot runs regardless.

## What was happening
`StatusItemController` bundled several concerns; one was the menu-bar strip render pipeline: `updateButtonImage` / `scheduleButtonImageUpdate` / `renderButtonImage`, the `pendingRenderTask` debounce, and the one-shot `withObservationTracking` re-arm.

## What this changes
- **`StatusItemImageUpdater`** owns that render loop. It needs only the `AppContainer` (layout `pinnedGroups` + `menuBarStyle`, `dataStore`) and an `apply: (NSImage) -> Void` closure to set the button image.
- `StatusItemController.init` creates it once the status button exists (`apply` = `self.statusItem.button?.image = …`) and calls `update()`; the pipeline methods + `pendingRenderTask` are gone from the controller.
- The debounce + `withObservationTracking` re-arm (the guard against the "menu bar disappears" refresh-storm) move intact — `withObservationTracking` is self-contained, so ownership doesn't change its semantics.

## Heads-up
- Not unit-testable (AppKit menu-bar rendering), so verified with an app run: the log reports `Status item ready (button: true)`, the strip renders on launch, providers refresh, no crash.
- The controller is still ~590 LOC; the residual is its genuine job (panel construction, status item + context menu, transparency, outside-click monitors, show/hide). The outside-click-guard dedup was already done in #868.

## Tests
- `swift test` — 883 tests, 0 failures. App launches and renders the strip.

---

> [!NOTE]
> **Low Risk**
> Behavior-preserving structural refactor of menu-bar rendering; no changes to panel, dismissal, or transparency logic.
> 
> **Overview**
> Moves the menu-bar strip render pipeline out of **`StatusItemController`** into a new **`StatusItemImageUpdater`** type. The controller no longer owns `updateButtonImage`, `scheduleButtonImageUpdate`, `renderButtonImage`, or `pendingRenderTask`; it holds an `imageUpdater` and kicks off rendering with `imageUpdater.update()` after setup.
> 
> **`StatusItemImageUpdater`** takes `AppContainer` plus an `apply: (NSImage) -> Void` closure. Init wires `apply` to `statusItem.button?.image = image` by capturing the local `NSStatusItem` (not `self`), avoiding retain cycles and optional workarounds before `super.init()`. Observation tracking, 50ms debounce, and `MenuBarContentBuilder` / `MenuBarStripRenderer` logic are unchanged—only relocated. A comment on `applyTransparency()` now points at `StatusItemImageUpdater.update()` for the one-shot re-arm pattern.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 84879dd4704e2974b2cbf4e1aa44f1f6e77afe5a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/OpenUsage/App/StatusItemController.swift
+++ b/Sources/OpenUsage/App/StatusItemController.swift
@@ -28,6 +28,8 @@ final class StatusItemController: NSObject {
     private let container: AppContainer
     private let updater: UpdaterController
     private let statusItem: NSStatusItem
+    /// Owns the menu-bar strip render loop (created in `init` once the status button exists).
+    private var imageUpdater: StatusItemImageUpdater?
     private let panel: MenuBarPanel
     private let hostingController: NSHostingController<AnyView>
     /// The panel's backdrop: an opaque tray by default, swapped to a behind-window vibrancy view when
@@ -97,7 +99,10 @@ final class StatusItemController: NSObject {
 
         configurePanel()
         configureStatusItem()
-        updateButtonImage()
+        imageUpdater = StatusItemImageUpdater(container: container) { [weak self] image in
+            self?.statusItem.button?.image = image
+        }
+        imageUpdater?.update()
         applyTransparency()
 
         appearanceObserver = NotificationCenter.default.addObserver(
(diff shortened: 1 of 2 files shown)


robinebers (commented):
Deep-review pass (DRY / SOLID / KISS focus). The "verbatim relocation" claim holds: I diffed the moved bodies token-by-token — `update()`/`schedule()`/`renderButtonImage()` match the old methods exactly, including the 50ms debounce, `Task.isCancelled` guard, `[weak self]` hops, and the icon fallback chain. The `withObservationTracking` closure still reads the same observable set through `container`, the retain topology is equivalent (controller strongly owns updater, updater→controller weak via the apply closure — no cycle, no lost updates), startup ordering is unchanged, and the `@MainActor` isolation is sound under strict concurrency. No blockers.

**Should fix before merge:**

1. **Stale doc comment** — `StatusItemController.swift:218`: `applyTransparency`'s doc still says "Mirrors `updateButtonImage`'s `withObservationTracking` re-arm", but `updateButtonImage` no longer exists anywhere in the codebase (this comment is the only remaining mention). Point it at `StatusItemImageUpdater.update()`.

**Nit:**

- The `imageUpdater` property doc says "created in `init` once the status button exists", but creation is unconditional — nothing checks the button (and doesn't need to; both old and new resolve `statusItem.button?` lazily at each apply). The optional `var` + `imageUpdater?.update()` exists only because `[weak self]` can't be captured before `super.init()`. Having the apply closure capture the `NSStatusItem` directly (`{ statusItem.button?.image = $0 }` — no self, no cycle, NSStatusItem doesn't retain the controller) would allow a non-optional `let` and delete both the `?`-call and the misleading comment.

---

PR #872, opened 2026-07-07
refactor(DashboardView): extract PanelHeightCoordinator (auto-fit computation)

robinebers:
**TL;DR** — Fourth and final god-object split. Pulls the popover auto-fit **height computation** out of `DashboardView` into a `PanelHeightCoordinator`, leaving the timing-sensitive animation mechanism untouched. Newly unit-tested. Stacked on #871.

> Top of the stack (#871 → #870 → #869 → #868). GitHub CI runs once the stack retargets to main; verified locally (build + 889 tests + **app launch**). Bugbot runs regardless.

## What was happening
`DashboardView` (~770 LOC) tangled the auto-fit/slide engine with the screen content. The auto-fit had two parts: the *measurement/target computation* (per-screen measured pieces → clamped ideal height) and the *animation* (`animatedHeight`, the slide, the `withAnimation` spring). The measurement was buried in the view and not unit-testable.

## What this changes
- **`PanelHeightCoordinator`** (`@Observable`, held as `@State`) owns the measured scroll-content/footer dictionaries, `recomposeIdeal`, and `target(for:)`/`clamped(_:)`. Geometry callbacks call `setScrollContent`/`setFooter`; the morph `onChange` reads `coordinator.measuredIdeal[…]`.
- The **animation stays in the view**: `animatedHeight`, `slideProgress`, `animatedSlideID`, `slideOffset`, and every `withAnimation` are unchanged — deliberately, so there's no risk to the delicate slide/morph timing. Only the deterministic measurement moved.

## Heads-up
- Scoped on purpose: I extracted the computation (safe, and now testable) rather than the animation clock. Lifting the `withAnimation`/`@State` timing into a separate object was the risky part with no way to verify slide/morph timing programmatically here — not worth it for an organizational change.
- The measurement is now covered by **6 new unit tests** it never had. App run confirms the popover still sizes correctly with no "Modifying state during view update".
- Nice pre-merge smoke test (not automatable from CI): open the popover and switch Dashboard ↔ Customize ↔ Settings to eyeball the slide+height morph. The animation code is unchanged, so this is a sanity check rather than a likely regression.

## Tests
- `swift test` — 889 tests (6 new `PanelHeightCoordinatorTests`), 0 failures. App launches, popover sizes correctly, no crash.

---

> [!NOTE]
> **Medium Risk**
> Touches live popover sizing and screen-switch morph targets; behavior is intended to be a pure move but regressions would show as wrong panel height during Dashboard ↔ Customize ↔ Settings transitions.
> 
> **Overview**
> Moves popover **auto-fit height math** out of `DashboardView` into a new `@Observable` **`PanelHeightCoordinator`**, held as `@State`. Per-screen scroll and footer measurements now flow through `setScrollContent` / `setFooter`; morph targets come from `measuredIdeal` and `target(for:)` instead of view-local dictionaries and helpers (`recomposeIdeal`, `targetHeight`, `clampedTarget`).
> 
> **Animation is unchanged in the view**: `animatedHeight`, slide progress, and all `withAnimation` springs still live in `DashboardView`—only the deterministic measurement and clamping moved.
> 
> Adds **`PanelHeightCoordinatorTests`** (six cases) for ideal composition (dashboard vs other screens, footer defaults, unset until content), clamp behavior via `MenuBarPopover.clampHeight`, and remeasurement when content grows.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9cf0c62de4396c3bab2ce23ccb35062b796afd08. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/OpenUsage/Views/DashboardView.swift
+++ b/Sources/OpenUsage/Views/DashboardView.swift
@@ -190,7 +190,7 @@ struct DashboardView: View {
                     // destination is usually mounted+measured by now (it mounted on the slideProgress=0
                     // render), so we morph to its ideal. If it ISN'T measured yet and the height was
                     // never established (animatedHeight still the 0 sentinel — e.g. opening Settings
-                    // straight from the status-item menu), we must NOT morph to clampedTarget(0), which
+                    // straight from the status-item menu), we must NOT morph to a clamped zero, which
                     // floors to minPanelHeight and wrongly shrinks the panel: leave the height alone and
                     // let the completion / measurement establish it once a real ideal lands.
                     let coTarget: CGFloat? = heightCoordinator.target(for: destination)
--- a/Sources/OpenUsage/Views/DashboardView.swift
+++ b/Sources/OpenUsage/Views/DashboardView.swift
@@ -31,12 +31,11 @@ struct DashboardView: View {
     /// Whether `animatedHeight` has been seeded for this open. Until then the first measurement (or a
     /// reopen) establishes it without animation; afterwards, changes spring.
     @State private var didEstablishHeight = false
-    /// Per-screen intrinsic heights, summed into `measuredIdeal`. Written only from geometry/preference
-    /// actions (which run after `body`), so they never trip "Modifying state during view update".
-    @State private var measuredScrollContent: [PopoverScreen: CGFloat] = [:]
-    @State private var measuredFooter: [PopoverScreen: CGFloat] = [:]
-    /// The window height each screen wants — top bar + footer + scroll content. The morph target.
-    @State private var measuredIdeal: [PopoverScreen: CGFloat] = [:]
+    /// Popover auto-fit height computation: per-screen measured pieces summed into each screen's clamped
+    /// morph target (`heightCoordinator.measuredIdeal` / `.target(for:)`). Written from the geometry
+    /// actions below. The animation itself — `animatedHeight`, the slide, the `withAnimation` spring —
+    /// stays in this view; the coordinator holds only the deterministic measurement.
+    @State private var heightCoordinator = PanelHeightCoordinator(topBarHeight: Self.topBarHeight)
     /// Horizontal screen-switch slide: 0 shows the outgoing screen, 1 the incoming one. Drives the
     /// page offset so the screens slide between modes on one spring.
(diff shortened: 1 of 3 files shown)


robinebers (commented):
Deep-review pass (DRY / SOLID / KISS focus). The extraction verifies as behavior-preserving: `recomposeIdeal` is a line-for-line move (top-bar 0 for dashboard / 44 injected otherwise, footer `?? 0`, unset-until-measured guard), `target(for:)` matches both old call shapes, and the SwiftUI wiring is sound — the coordinator is an `@Observable` held in `@State`, `onChange(of: heightCoordinator.measuredIdeal[layout.screen])` still registers observation during body evaluation, writes only happen from geometry/preference callbacks, and the old `@State` dicts are fully gone (no double source of truth). Marking the measurement dicts `@ObservationIgnored` is a mild improvement (drops spurious invalidations). The split boundary — deterministic measurement out, animation clock stays — is the right call. No blockers.

**Should fix before merge:**

1. **`testTargetIsNilUntilMeasuredThenClamps` never clamps anything** (`PanelHeightCoordinatorTests.swift:43`). It asserts only the identity fallback when `MenuBarPopover.clampHeight` is nil — the actual clamp behavior (the one interesting edge) is unpinned. Install a hook (e.g. `{ min(max($0, 400), 600) }`), assert both edges, and save/restore the global — the sibling `PanelHeightBridgeTests` already models this reset pattern. As written the test also silently depends on the process-global being nil, so any future test that installs a hook makes this suite order-dependent.

**Nits:**

- `DashboardView.swift:193`: comment references `clampedTarget(0)`, which no longer exists in this file (it became the coordinator's `clamped`).
- `PanelHeightCoordinator.clamped(_:)` is non-private with a doc claiming "e.g. tests", but nothing outside the type calls it — make it private or actually use it from the clamp test above.
- `import SwiftUI` + `import Observation` is overweight for a deliberately view-free type: no SwiftUI API is used, and SwiftUI re-exports Observation anyway. `CoreGraphics` + `Observation` states the real dependency.

(Not flagged: DashboardView is still ~748 LOC vs the ~500 guideline, but this PR reduces it and the remainder is pre-existing.)

---

PR #878, opened 2026-07-08
Stop tests leaking UserDefaults suite domains

robinebers:
## TL;DR

The test suite created unique `UserDefaults(suiteName:)` domains but usually cleared them only before a test, never afterward, so repeated `swift test` runs accumulated thousands of test preference plists in `~/Library/Preferences`. This routes every leaking scratch suite through one shared helper that removes its data in teardown. Test infrastructure only — no app behavior change.

## What was happening

- Test helpers built suites with fresh UUIDs, called `removePersistentDomain` once at the top for a clean slate, and returned them.
- Nothing removed those domains when the test finished, so each run left newly written preferences behind.
- A few files already used a correct `defer { removePersistentDomain }` pattern; those remain unchanged.

## What this changes

- Adds `XCTestCase.makeScratchDefaults(suiteName:)` in `TestSupport.swift`. It pre-clears the suite and registers an `addTeardownBlock` that removes its persistent domain after the test.
- Routes all 28 leaking call sites across 27 test files through that helper, including the newer `LayoutBootstrapTests` and `LayoutPersistenceTests` added after this branch was first created.
- Keeps the existing XCTest class hierarchy unchanged and continues returning plain `UserDefaults` instances to test bodies.

## Heads-up

- **macOS 27 caveat:** on macOS ≤ 26 (including the `macos-26` CI runner), `removePersistentDomain` removes the suite plist. On macOS 27 beta, `cfprefsd` recreates an empty 42-byte dictionary for touched suites when the test process exits. The teardown still removes all test data, but a post-run filesystem cleanup is required to remove those empty shells.
- A complete one-time cleanup must include the non-`OpenUsageTests` prefixes used by Copilot, notification, and snapshot-cache tests:

```sh
find "$HOME/Library/Preferences" -maxdepth 1 -type f \( \
  -name 'OpenUsageTests*.plist' -o \
  -name 'CopilotProviderTests-*.plist' -o \
  -name 'WidgetDataStoreNotificationTests*.plist' -o \
  -name 'providerSnapshotCache.test.*.plist' \
\) -delete
```

## Tests

- `swift test --filter 'Layout(Bootstrap|Persistence)Tests'` — 5 passed.
- `swift test` — 953 passed, 3 skipped, 0 failures.
- Verified from a purged clean slate on macOS 27: the full suite created 216 plist shells, all 42-byte empty dictionaries; zero matching files retained test data.
- Removed the pre-existing 5,623-file local backlog and the post-verification empty shells; both matching filesystem files and `defaults domains` were zero afterward.

### DIFF
--- a/Tests/OpenUsageTests/AntigravityLayoutTests.swift
+++ b/Tests/OpenUsageTests/AntigravityLayoutTests.swift
@@ -101,10 +101,7 @@ final class AntigravityLayoutTests: XCTestCase {
     }
 
     private func makeDefaults(_ name: String) -> UserDefaults {
-        let suiteName = "OpenUsageTests.AntigravityLayout.\(name).\(UUID().uuidString)"
-        let defaults = UserDefaults(suiteName: suiteName)!
-        defaults.removePersistentDomain(forName: suiteName)
-        return defaults
+        makeScratchDefaults(suiteName: "OpenUsageTests.AntigravityLayout.\(name).\(UUID().uuidString)")
     }
 
     private func saveStored<T: Encodable>(_ value: T, forKey key: String, in defaults: UserDefaults) {
--- a/Tests/OpenUsageTests/CopilotProviderTests.swift
+++ b/Tests/OpenUsageTests/CopilotProviderTests.swift
@@ -567,10 +567,7 @@ final class CopilotProviderTests: XCTestCase {
     }
 
     private func freshDefaults() -> UserDefaults {
-        let suiteName = "CopilotProviderTests-\(UUID().uuidString)"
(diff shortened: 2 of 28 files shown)


robinebers:
I found one issue that should be fixed before merging.

This covered every leaking test when the branch was created, but two layout test groups were added to `main` afterward. Their temporary preferences still get cleared only before the test, not after it. That means a full test run would continue leaving those two sets of saved data behind, while the PR says it stops the leak everywhere.

Please update the branch from `main` and route both newer layout helpers through the shared cleanup helper as well. The shared helper and the rest of this cleanup look sound.

PR #921, opened 2026-07-09
Keep automatic refresh cadence stable

robinebers:
## TL;DR

Makes `AppContainer` the single automatic refresh owner, preserves the five-minute timer deadline across early provider-enablement wakes, and gives the footer the scheduler-owned deadline so manual or cache-hit refreshes cannot move the displayed countdown.

## What was happening

- `AppContainer` started the launch refresh while the eagerly hosted dashboard started another one-shot refresh.
- Provider guards prevented duplicate requests, but the losing batch could complete with skipped outcomes and let notification evaluation inspect stale state.
- Every early enablement wake restarted a full interval after its cache-aware pass, so a wake just before expiry could stretch live data updates toward ten minutes.
- The footer derived its countdown from the most recent batch completion, even when that batch was manual or an early cache hit and did not move the real timer.

## What this changes

- Removes the view-owned initial refresh and keeps one process-lifetime automatic loop in `AppContainer`.
- Extracts a small scheduler seam with explicit refresh, notification, telemetry, and wait ordering.
- Stops cancellation during refresh or notification delivery before telemetry, deadline publication, or another wait.
- Distinguishes enablement wakes from scheduled timer wakes.
- Preserves a monotonic deadline across early wakes and immediately runs the due scheduled pass if an early pass crosses that deadline.
- Publishes `nextAutomaticRefreshAt` from the scheduler only when cadence advances.
- Makes the footer read that deadline; manual refreshes and early wakes no longer reset its countdown.
- Updates architecture, refresh behavior, research notes, and in-flight ownership comments.

## Heads-up

- Manual and timer requests still use the existing per-provider in-flight guard. Whole-batch forced-refresh coalescing is isolated in PR #928.
- A merge-tree check with PR #928 is clean and the behaviors are compatible.
- No visual design change; screenshots are not applicable.

## Tests

- Eleven focused scheduler, wake-signal, countdown, cancellation, and manual-refresh lifecycle tests passed.
- Added deterministic regressions for a wake four minutes into cadence, a wake pass crossing expiry, and a manual refresh preserving the deadline.
- Full suite: 969 XCTest cases passed with 3 expected skips, plus 3 Swift Testing cases.
- `git diff --check` passed.
- `script/build_and_run.sh verify` completed the release build, signed app staging, launch, and process verification.
- Independent exact-commit and combined-tree review found no remaining lifecycle issue.

---

> [!NOTE]
> **Medium Risk**
> Touches core refresh scheduling and timing behavior that affects all providers and the footer UX; changes are well-tested but incorrect deadline logic could delay live updates or confuse users.
> 
> **Overview**
> **Automatic refresh** is consolidated in `AppContainer` via a new **`PeriodicRefreshLoop`**: refresh → notifications → telemetry → wait, with cancellation guards so half-finished cycles do not publish deadlines or run post-refresh work.
> 
> **`DashboardView`** no longer runs a one-shot `refreshAll()` on mount, so opening the popover cannot compete with the background loop.
> 
> **`RefreshWakeSignal`** now returns **enablement vs timer** triggers (`wait(timeout:)`). Early provider-enablement passes **keep the existing monotonic deadline** instead of sleeping another full interval; if an early pass crosses expiry, the next wait uses **zero** so the scheduled pass runs immediately.
> 
> **`WidgetDataStore`** drops batch-end `lastRefreshAt` stamping and exposes **`nextAutomaticRefreshAt`**, set only when the scheduler starts a new cadence interval—manual ⌘R refreshes do not move it.
> 
> **`PopoverFooter`** formats “Next update in …” from that deadline via **`countdownText`**.
> 
> Docs and **unit tests** cover loop ordering, wake buffering, early-wake timing, deadline crossing, and manual refresh preserving the deadline.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c3e4a788a9598f3975b1e54dc01ca697460ba84c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/OpenUsage/App/AppContainer.swift
+++ b/Sources/OpenUsage/App/AppContainer.swift
@@ -168,10 +168,10 @@ final class AppContainer {
         FirstRunSeeder.reseed(providers: providers, enablement: enablement)
     }
 
-    /// Drives live updates: refresh on launch, then again every refresh interval. Each pass honors the
-    /// cache, so it only hits the network once a snapshot has actually expired. `@Observable` propagates
-    /// the resulting snapshot changes to the menu-bar label and any open widgets, so the UI refreshes on
-    /// its own instead of only when the popover opens.
+    /// Sole owner of automatic updates: refresh on launch, then again every refresh interval. Each pass
+    /// honors the cache, so it only hits the network once a snapshot has actually expired. `@Observable`
+    /// propagates the resulting snapshot changes to the menu-bar label and any open widgets; mounting or
+    /// opening the popover never starts a competing refresh pass.
     ///
     /// Between passes the loop sleeps via `RefreshWakeSignal`, which wakes it early when the user
     /// enables/disables a provider so a newly-enabled provider is fetched promptly instead of waiting out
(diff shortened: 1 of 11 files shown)


robinebers:
Superseded in part by #950. The proven duplicate DashboardView launch refresh was retained, but the new scheduler/deadline abstraction and 237 lines of scheduler tests were deliberately dropped because they add more state than the clarified optimization goal justifies. The original branch remains preserved for a focused cadence bug if needed; nothing was merged.

PR #925, opened 2026-07-09
Preserve Cursor token refresh failure semantics

robinebers:
## TL;DR

Stops Cursor token-refresh network, server, rate-limit, and malformed-response failures from being mislabeled as an expired login, while preserving the existing safe fallback to an access token the usage endpoint may still accept.

## What was happening

- Refresh transport failures were untyped, while refresh 5xx, 429, malformed JSON, and missing-token responses collapsed to `nil`.
- During 401 recovery, that `nil` became “Token expired,” encouraging a re-login that cannot fix infrastructure or response-schema failures.
- A transport failure on the retried usage request had a distinct friendly message but was counted as telemetry category “other” instead of “network.”
- Proactive refresh failures with an existing access token were swallowed without a diagnostic.
- The token-persistence failure log named SQLite even when the selected credential source was Keychain.

## What this changes

- Maps refresh transport failures to the network category and preserves HTTP status semantics for 429, 4xx, and 5xx responses.
- Treats malformed JSON, malformed 400/401 bodies, and successful responses without a usable access token as invalid responses.
- Reserves session/token-expired errors for Cursor explicit `shouldLogout` boolean semantics.
- Keeps the existing proactive fallback to an access token that may remain server-usable, but logs that fallback without tokens or raw errors.
- Classifies transport failure after a successful refresh/retry as network while retaining its specific user-facing message.
- Makes the persistence diagnostic accurate for either SQLite or Keychain without exposing the credential or underlying SQL error.
- Updates Cursor troubleshooting guidance for the distinct token-refresh failure classes.

## Heads-up

- PR #914 independently corrects the pre-existing stale sentence about whether Cursor CSV export is fetched.
- PR #929 owns CSV export validation and fail-loud diagnostics; other optional endpoint diagnostics are intentionally kept out of this token-refresh PR.

## Tests

- Cursor provider suite: 6 tests passed; all Cursor-focused tests passed during independent review.
- Added coverage for refresh-only network failure; 429, 503, malformed/missing-token responses; explicit logout/token expiry; proactive fallback; and 401 → refresh → retried transport failure.
- Full suite: 967 XCTest cases passed, 3 expected skips; 3 Swift Testing cases passed.
- Release build packaged, code-signed, and launched with `script/build_and_run.sh verify`.
- `git diff --check` passed and two independent review passes found no remaining token-refresh issue.

## Screenshots

Refresh infrastructure failures now retain actionable guidance instead of appearing as expired authentication.

Cursor refresh failure notice

### DIFF
--- a/Sources/OpenUsage/Providers/Cursor/CursorProvider.swift
+++ b/Sources/OpenUsage/Providers/Cursor/CursorProvider.swift
@@ -88,6 +88,10 @@ final class CursorProvider: ProviderRuntime {
                 if accessToken == nil {
                     throw error
                 }
+                // A near-expiry JWT may still be accepted by the usage endpoint, so retain the existing
+                // fallback. Keep the swallowed refresh failure diagnosable without interpolating an
+                // error that could contain request details or credentials.
+                AppLog.warn(LogTag.auth("cursor"), "proactive token refresh failed; trying the existing access token")
             }
         }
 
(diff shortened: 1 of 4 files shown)


robinebers:
Closing after the code-quality portfolio re-scope. The Cursor refresh-error classification is useful but medium/low severity and carries substantially more test churn than production simplification. It also overlaps the narrower Cursor boundary-integrity replacement now being prepared. The branch remains preserved.

PR #926, opened 2026-07-09
Preserve and normalize saved layout identifiers

robinebers:
## TL;DR

Normalizes persisted dashboard layouts at startup and keeps their saved identifiers stable across every layout mutation. This repairs malformed local state without erasing identifiers written by a newer OpenUsage version.

## What was happening

- Persisted provider, metric, expansion, menu-bar, and placement arrays were trusted in several different forms.
- Duplicate known descriptors and UUIDs could survive bootstrap, while ordinary layout changes could accidentally discard unknown identifiers created by a newer app version.
- Startup repairs and the seeded marker were not coordinated as one crash-safe persistence operation.

## What this changes

- Centralizes bootstrap normalization and the persistence projection for all eight identifier-bearing layout fields.
- Repairs duplicate known descriptors and UUIDs while retaining opaque future identifiers through mutations, resets, and downgrade/upgrade cycles.
- Restricts direct mutation of normalized ordering state and simplifies the live ordering path.
- Writes repaired layout state before the seeded marker so an interrupted startup can recover on the next launch.
- Updates the architecture and settings documentation to describe normalization and forward compatibility.

## Heads-up

- This intentionally preserves unknown persisted identifiers; the running app ignores them until a version that understands them is installed.
- There is no visual design change, so screenshots are not applicable.

## Tests

- Full Swift suite: 967 XCTest cases passed, 3 skipped; 3 Swift Testing cases passed.
- Focused layout suites: 79 tests passed.
- Added forward-compatibility, duplicate-descriptor, duplicate-UUID, interrupted-write, and mutation regressions.
- Release build packaged, code-signed, and launched with `script/build_and_run.sh verify`.

---

> [!NOTE]
> **Medium Risk**
> Changes how all layout UserDefaults are loaded, repaired, and written at startup and on every customization save; regressions could alter user dashboards or lose forward-compatible data, though coverage is extensive.
> 
> **Overview**
> **Layout startup** now repairs saved dashboard state in one place: `LayoutBootstrap` builds separate **runtime** and **persistence** projections for all eight identifier-bearing fields (placement, provider/metric order, pins, expand/caret membership, seeded defaults). Unknown IDs from a newer app version stay in storage at their original positions but are hidden at runtime; duplicates and UUID collisions among known widgets are fixed without dropping opaque entries.
> 
> **Persistence** gains `LayoutPersistencePlan` / `activate(_:registry:)`, ordered crash-safe repair writes (seed marker last), and merge-on-save so later edits only replace known slots—an older build cannot wipe future provider/metric settings on reset or customize. Loads distinguish **missing** keys from **present but malformed** data so corruption is repaired instead of treated as a fresh install.
> 
> **`LayoutStore`** applies the plan at init instead of ad-hoc persist flags; `placed`, `providerOrder`, and `metricOrderByProvider` are `private(set)`; live ordering no longer re-normalizes on every read. Docs and tests (including downgrade/upgrade round-trips) describe and lock in forward compatibility.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3a30ae991ef23a62e10997893ba36ac8c696e5b9. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


robinebers:
Closing after the code-quality portfolio re-scope. The forward-compatibility edge case is valid, but this implementation adds dual projections and per-field merge logic across every persisted layout field—688 production lines of churn in a hot path. That review and complexity cost is not justified by the clarified KISS goal. The branch remains preserved.

PR #930, opened 2026-07-09
Log Cursor optional endpoint failures

robinebers:
## TL;DR

Makes Cursor optional enrichment failures visible without letting nonessential plan, credit, prepaid-balance, or request-fallback endpoints discard usable primary usage, and rejects malformed financial metadata before it can produce an incorrect balance.

## What was happening

- Several Cursor enrichment calls used `try?` or returned `nil` for every failure shape.
- Transport, request preparation, non-success status, invalid JSON, and malformed metadata were indistinguishable from absent optional data.
- The request-based compatibility fallback could fail silently before the primary mapping error surfaced.
- Credit-grant fields defaulted malformed totals or usage to zero, which could overstate remaining credit.
- Prepaid-balance HTTP parsing lived in the mapper instead of at the provider boundary.

## What this changes

- Adds one provider helper for optional JSON boundaries, removing repeated transport, status, and schema handling.
- Records fixed, credential-free reasons for plan, credit-grant, and prepaid-balance failures.
- Validates and trims plan metadata while preserving request-based fallback when metadata is unusable.
- Requires a real credit-grant flag and, for active grants, finite positive totals plus nonnegative usage before mapping financial data.
- Logs generic request-fallback failure before continuing to the primary mapping result.
- Passes already-decoded prepaid data into the mapper, keeping I/O validation at the provider boundary.
- Keeps optional failures nonfatal when primary Cursor usage is usable.
- Updates Cursor troubleshooting with a dedicated optional-lookup entry.

## Tests

- Six boundary tests cover transport, request preparation, HTTP status, invalid JSON, plan/prepaid/credit metadata, fallback behavior, log visibility, token privacy, and primary-usage preservation.
- Includes a regression proving malformed used-credit metadata cannot create a bogus Credits row.
- Full suite: 968 XCTest cases passed with 3 expected skips, plus 3 Swift Testing cases.
- `git diff --check` passed.
- `script/build_and_run.sh verify` completed the release build, signed app staging, launch, and process verification.

No visual layout changes; screenshots are not applicable.

### DIFF
--- a/Sources/OpenUsage/Providers/Cursor/CursorProvider.swift
+++ b/Sources/OpenUsage/Providers/Cursor/CursorProvider.swift
@@ -123,18 +123,20 @@ final class CursorProvider: ProviderRuntime {
         }
 
         if shouldTryGenericRequestFallback(usage: usage) {
-            if let mapped = try? await requestBasedResult(
-                accessToken: currentToken,
-                planName: planName,
-                unavailableMessage: "Cursor request-based usage data unavailable. Try again later."
-            ) {
+            do {
+                let mapped = try await requestBasedResult(
+                    accessToken: currentToken,
+                    planName: planName,
+                    unavailableMessage: "Cursor request-based usage data unavailable. Try again later."
+                )
                 return snapshot(mapped)
+            } catch {
+                AppLog.warn(LogTag.plugin("cursor"), "optional request-based usage fallback failed")
             }
         }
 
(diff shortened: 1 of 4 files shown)


robinebers:
Superseded by #948, which retains the single optional-endpoint validation helper and prepaid boundary checks alongside the related Cursor CSV integrity work. #948 is under 1,000 gross lines, fully green, and avoids overlapping Cursor PRs. This branch remains preserved; nothing was merged.

PR #942, opened 2026-07-10
Surface CLI provider credential storage failures

robinebers:
## TL;DR

Distinguish genuinely missing credentials from unreadable or malformed local credential stores for Codex, Copilot, and Antigravity. Valid later sources still win, while a real boundary failure now reaches the user with actionable guidance instead of being mislabeled as “not logged in.”

## What was happening

- Codex auth files and Keychain reads collapsed missing, unreadable, malformed, and tokenless data into the same empty result.
- Copilot editor, GitHub CLI, and Keychain failures could disappear as “not logged in,” and one broken higher-priority file could hide a valid later source.
- Antigravity Keychain failures were treated as missing credentials; malformed structured values could fall through as raw bearer tokens.
- One-shot provider detection could leave a provider disabled when credentials probably existed but could not be inspected.
- Troubleshooting docs did not explain these repairable credential-store failures or the intentional best-effort exceptions.

## What this changes

- Adds typed unreadable/malformed credential outcomes and telemetry categories for all three providers.
- Makes provider detection conservative for indeterminate credential boundaries while keeping proven absence disabled.
- Preserves ordered fallback: a valid later Codex file, Copilot source, Keychain entry, or Antigravity refreshed-token cache can recover from an earlier broken source.
- Keeps valid Codex API-key-only guidance ahead of a broken sibling while still allowing later OAuth candidates to serve subscription usage.
- Restricts Copilot parsing to GitHub.com credentials; valid GitHub Enterprise-only configuration remains compatible absence and falls through safely.
- Treats Antigravity’s app-owned refreshed-token cache as best effort: expired entries are normal misses and corrupt/unreadable entries are logged and ignored.
- Documents the new errors, recovery steps, fallback behavior, and cache/enterprise exceptions.

## Heads-up

- This PR is stacked on #941; review the commit on top of `codex/surface-grok-devin-credential-errors`, and rebase it after #941 lands.
- #915 overlaps the Codex fallback switch. When rebasing, preserve `credentialStoreUnreadable` as a terminal load error while retaining API-key fallback.
- No visual changes.

## Tests

- `swift test --filter 'CredentialBoundaryTests'` — 30 tests passed.
- `swift test` — 1,009 XCTest tests passed (3 skipped), plus 3 Swift Testing tests.
- `script/build_and_run.sh verify` — release build, signed bundle staging, launch verification, and post-verification process cleanup passed.
[...]

PR #947, opened 2026-07-10
Bind Antigravity cached tokens to the current login

robinebers:
## TL;DR

Bind OpenUsage's short-lived Antigravity access-token cache to the Keychain login that produced it, so a logout, account switch, or unreadable Keychain can never reuse another account's cached quota credential. This is stacked on #942.

## What was happening

- A refreshed Antigravity access token was cached with only its expiry.
- That app-owned cache was treated as an independent credential when the Keychain login was absent or unreadable.
- A still-valid token could therefore fetch and display the previous account's quota after logout or an account switch.
- Cache and Keychain parsing tests were mixed into the already-large provider test file.

## What this changes

- Store a SHA-256 fingerprint of the source refresh credential beside each newly cached access token; the refresh credential itself is never copied.
- Reuse a cache entry only when the current Keychain login is present, readable, and has the matching fingerprint.
- Discard logout, account-mismatch, legacy-unbound, expired, empty, and malformed cache entries. A transient Keychain read failure keeps the cache for recovery but never uses it.
- Keep `hasLocalCredentials()` grounded in Antigravity's Keychain login instead of the derivative cache.
- Move refreshed-token persistence off the main actor.
- Add account-switch, logout, Keychain-failure, fingerprint, migration, malformed, unreadable, and expiry regression coverage; split token/cache tests into a dedicated file.
- Document the credential-cache behavior in both the Antigravity provider guide and privacy documentation.

## Heads-up

- This PR is stacked on #942 and should be retargeted to that PR's eventual base after #942 lands.
- Existing unbound cache files are deliberately invalidated once, which may cause one extra OAuth refresh.
- No visual behavior changes.

## Tests

- `swift test --filter 'Antigravity(TokenCache|CredentialBoundary|Provider)Tests'` — 43 passed.
- `swift test` — 1,015 XCTest tests passed, 3 skipped; 3 Swift Testing tests passed.
- `script/build_and_run.sh verify` — release build, signed app staging, launch, and running-process verification passed.

---

> [!NOTE]
> **Medium Risk**
> Changes Antigravity auth and local token caching paths where a wrong reuse could show another account’s quota; behavior is tightened with tests and docs, but credential handling is security-sensitive.
> 
> **Overview**
> **Antigravity’s OpenUsage-refreshed access-token cache is no longer treated as a standalone credential.** New writes store a SHA-256 fingerprint of the Keychain refresh token (not the refresh token itself). `loadCachedToken(matching:)` only returns a cached access token when the current Keychain login matches that fingerprint; logout, account switch, legacy unbound entries, expiry, and bad data trigger `discardCachedToken()` instead of silent reuse.
> 
> **Provider behavior shifts accordingly:** `hasLocalCredentials()` keys off Keychain login only, not the derivative cache. Cloud Code probing requires a readable Keychain token before using cache; proven logout clears the cache; Keychain read failures throw without using or deleting the cache for recovery. OAuth refresh caching runs off the main actor and passes the source refresh token when persisting.
> 
> Tests move token/cache cases into `AntigravityTokenCacheTests` and expand boundary coverage (account switch, logout, Keychain failure). **docs/privacy.md** and **docs/providers/antigravity.md** document the bound cache. Existing unbound cache files invalidate once (may cause one extra OAuth refresh).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e3e22959476ad21913ccb6a102e177c5bef0bcd0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


robinebers:
Superseded by #954, now stacked directly on #910 instead of the broad credential-error chain. The replacement preserves the complete account-isolation invariant: SHA-256 binding to the current refresh credential, matching-cache happy path, logout/account-switch/legacy/expired/malformed cleanup, zero stale-token requests during Keychain failure, cache retention for recovery, and off-main cache I/O. It avoids the test-file move and unrelated provider state machines, passes 975 XCTest cases plus 3 Swift Testing cases, and has zero review threads with all available checks green. I am closing this stacked version; its branch and history remain available.

PR #941, opened 2026-07-10
Surface Grok and Devin credential storage failures

robinebers:
## TL;DR

Distinguish missing Grok and Devin credentials from present-but-unreadable or malformed local storage, so setup detection no longer silently leaves affected providers off and refresh shows a repairable error instead of “not logged in.”

## What was happening

- File, SQLite, parse, and absence outcomes were collapsed with `try?` and optional loads.
- A permissions, encoding, database-lock, schema, or malformed-payload failure could be reported as an ordinary missing login.
- New-provider detection probes each provider once, so a transient credential-store failure could leave that provider disabled permanently.
- A read-only SQLite probe could open a missing path and create an empty database.

## What this changes

- Adds boundary helpers where `nil` means proven absence and real file/SQLite failures throw.
- Opens existing SQLite credential databases read-only and skips process launch when the database is absent.
- Gives Grok and Devin friendly typed errors for unreadable and malformed local credential storage, with privacy-safe source-only logging.
- Keeps ordered fallback behavior: any valid sibling credential wins; a boundary error surfaces only when no usable source loads.
- Makes credential detection conservative on an indeterminate store so normal refresh can explain the problem to the user.
- Adds a `credential_access` telemetry category and documents the provider/auth-store contract.

## Heads-up

- This PR intentionally stays based on `main`. When integrating with #927, retain that PR’s typed Devin remote-error precedence and regression tests; the credential-boundary and remote-response fixes are complementary.
- Other providers still using lossy optional loads are being migrated in follow-up PRs; the behavior documentation is scoped accordingly.

## Tests

- 26/26 focused file, SQLite, Grok, and Devin boundary tests passed.
- Full suite passed: 979 XCTest tests, 3 skipped, plus 3 Swift Testing tests.
- Release build staged, signed, launched, and confirmed running; the test app was stopped afterward.
- Independent final diff review found no scoped Grok, file, SQLite, or concurrency defect.
- `git diff --check` passed.

## Screenshots

- No component styling or layout changes; the new failures reuse the existing provider warning presentation.

---

> [!NOTE]
> **Medium Risk**
> Changes first-run/new-provider enablement and local credential reads for Grok and Devin; behavior is conservative on probe failures but could enable providers that previously stayed off when stores were broken.
> 
> **Overview**
[...]

PR #954, opened 2026-07-10
Bind credential caches to verified local state

robinebers:
## TL;DR

Makes local credential probes side-effect free and binds OpenUsage's Antigravity access-token cache to the Keychain login that produced it. Logout, account switches, malformed credentials, and unavailable Keychain access can no longer reuse or send a previous account's cached token.

## What was happening

- A read-only SQLite credential probe could create an empty database when the provider database was absent.
- File absence and file read failures were collapsed by exists-then-read call sites.
- Antigravity treated its derived access-token cache as an independent credential.
- A still-valid cache could therefore query the previous account after logout or an account switch.
- Keychain read failures looked like logout, and malformed structured values could be sent as bearer tokens.

## What this changes

- Adds one file-read boundary where `nil` means proven absence and other failures still throw.
- Skips sqlite3 entirely for a missing database and opens existing credential databases with `-readonly`.
- Stores only a SHA-256 fingerprint of the source Antigravity refresh credential beside a cached access token.
- Reuses the cache only for the matching, readable Keychain login; legacy, mismatched, expired, empty, and malformed entries are discarded.
- Removes the derivative cache as soon as detection or refresh proves logout, but retains and never uses it during a transient Keychain read failure.
- Surfaces unreadable and malformed Antigravity Keychain data through friendly categorized errors.
- Keeps cache reads and writes off the main actor and documents the privacy and recovery behavior.

## Heads-up

- This is stacked on #910 so every cache write also gets its atomic owner-only file semantics.
- It replaces the high-value credential/cache subset of #941, #942, and #947 without their broader provider state-machine expansion.
- The review diff is 523 gross lines: 219 production lines, 295 test lines, and 9 documentation lines.
- Existing unbound Antigravity cache files are invalidated once and may cause one extra OAuth refresh.

## Tests

- Focused credential/cache integrity suite: 11 passed.
- Full suite: 976 XCTest cases passed, 3 expected skips; 3 Swift Testing cases passed.
- Release build, app staging, Apple Development signing, and strict codesign verification passed.
- The exact staged executable launched as PID 23044 and stopped cooperatively; unrelated same-name processes were ignored.
- `git diff --check` passed.

No visual behavior changed; screenshots are not applicable.

---

> [!NOTE]
> **High Risk**
> Changes authentication and local credential handling for Antigravity, including when cached tokens may leave the machine; incorrect binding logic could break sign-in or briefly reuse the wrong account until cache invalidation.
> 
> **Overview**
> **Antigravity’s derived access-token cache is no longer treated as its own credential.** The refreshed-token file stores a **SHA-256 fingerprint** of the current Keychain refresh token and is reused only when that login is present, readable, and unexpired; logout, account switches, legacy unbound caches, and bad entries are **discarded** and never sent on the wire. **Keychain reads now throw** distinct errors for unreadable vs malformed data (mapped to new **`credential_access`** / **`auth_invalid`** categories), and **`hasLocalCredentials`** keys off Keychain only—clearing the cache on proven logout but keeping it (unused) during transient Keychain lock failures.
> 
> **Credential I/O is tightened platform-wide for probes:** `readTextIfPresent` separates missing files from read failures, SQLite credential queries **skip missing DBs** and open existing ones **`-readonly`**, and a **throwing `loadOffMainActor`** supports propagating those failures. Keychain parsing **rejects broken `{`/`[` payloads** as bearer tokens. Docs and a focused integrity test suite cover logout, account switch, and cache behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dc939e2b2fc500f3654cd1024cf289e909e5579a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/OpenUsage/Providers/Antigravity/AntigravityAuthStore.swift
+++ b/Sources/OpenUsage/Providers/Antigravity/AntigravityAuthStore.swift
@@ -1,3 +1,4 @@
+import CryptoKit
 import Foundation
 
 /// Credentials Antigravity already has on the machine. On current builds the OAuth tokens live in the
@@ -34,11 +35,21 @@ struct AntigravityAuthStore: Sendable {
     }
 
     /// Blocking keychain read — call off the main actor.
-    func loadKeychainToken() -> AntigravityKeychainToken? {
-        guard let raw = try? keychain.readGenericPassword(service: Self.keychainService, account: Self.keychainAccount),
-              let token = Self.extractToken(fromKeychainRaw: raw)
-        else {
-            return nil
+    func loadKeychainToken() throws -> AntigravityKeychainToken? {
+        let raw: String?
+        do {
+            raw = try keychain.readGenericPassword(
+                service: Self.keychainService,
+                account: Self.keychainAccount
+            )
+        } catch {
+            AppLog.error(LogTag.auth("antigravity"), "keychain credential read failed")
+            throw AntigravityError.credentialStoreUnreadable
+        }
+        guard let raw else { return nil }
+        guard let token = Self.extractToken(fromKeychainRaw: raw) else {
+            AppLog.error(LogTag.auth("antigravity"), "keychain credential is malformed")
+            throw AntigravityError.invalidCredentialData
         }
         return token
     }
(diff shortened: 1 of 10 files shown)


robinebers:
Closed when base branch `codex/secure-credential-files` (#910) was deleted on merge. Rebased onto `main` and landed as #961.

PR #965, opened 2026-07-11
feat: add Claude multi-account support

### LINKED ISSUE #402: Support multiple accounts per provider / platform
## Problem

OpenUsage currently appears to model usage primarily as:

```text
provider -> usage
```

This works for simple setups, but it breaks down when a user has multiple accounts on the same platform/provider.

Examples:

- multiple OpenAI / Codex accounts
- multiple Google / Gemini accounts
- personal + work accounts on the same platform
- multiple accounts behind the same gateway/proxy provider

In these cases, the important unit is often not just the provider, but:

```text
provider × account × quota window
```

## Why this matters

For subscription-based AI tools, accounts may have separate quota windows, reset times, rate-limit states, or even different plan sizes.

A single provider-level usage bar can be misleading:

- showing only the minimum remaining quota makes the whole provider look exhausted
- averaging account usage is wrong when accounts have different plan sizes
- merging accounts hides which account is constrained
- only showing one detected account makes multi-account setups hard to monitor

For example:

```text
OpenAI / Codex
  - Personal account
  - Work account
  - Backup account
```

These may share a provider implementation but should still be distinguishable as account instances.

## Suggested design

Consider adding first-class support for provider account instances.

Possible model:

```ts
ProviderInstance {
  providerId: string
  accountId: string
  displayName: string
  plan?: string
  authSource?: string
  usageLines: UsageLine[]
}
```

UI could show:

```text
OpenAI / Codex
  Personal    80% / 60%
  Work        35% / 90%
  Backup       — / —
```

For providers where multiple accounts act as a shared scheduling pool, the UI could still show account rows but avoid a misleading aggregate percentage unless account capacity/plan weights are known.

## Desired behavior

- allow multiple accounts under the same provider
- preserve account display names / aliases
- show usage per account when available
- support selecting an active account where relevant
- gracefully hide removed accounts
- support mixed known/unknown quota data
- avoid forcing all accounts into a single provider-level percentage

## Related

This is related to but broader than #280, which reports a Copilot-specific multi-account detection problem.

## Summary

Please consider changing the data model from:

```text
provider -> usage
```

to:

```text
provider -> account instances -> usage
```

This would make OpenUsage much more useful for users with personal/work/backup accounts on the same AI platform.

QuadDepo:
## Approved issue

Fixes #402 (not approved but reopend? partially fixes it; only tackles Claude accounts)

## TL;DR

Ability to have multiple Claude accounts and give them an appropriate name.

## What was happening

I got annoyed by the fact that I couldn't see my work usage, but only personal. 

## What this changes

Add support for multiple Claude code accounts by checking `~/.claude*` to find additional `.credentials.json` or getting them from the keychain with the `claude Code-credentials` prefix.

## Heads-up

- Currently only covers Claude account
- Opening the status page is not account bound (not a regression, depends on what logged-in account on browser is)

## Tests

I have verified the changes locally using a build, as well as written the correct tests where needed.

## Screenshots

customize
with-without-name

### DIFF
--- a/Sources/OpenUsage/App/AppContainer.swift
+++ b/Sources/OpenUsage/App/AppContainer.swift
@@ -29,6 +29,9 @@ final class AppContainer {
     /// One-time onboarding state (the first-run Customize hint card). Only ever marked pending by
     /// `FirstRunSeeder` on a fresh install, so existing installs never see the card.
     let onboarding: OnboardingStore
+    /// Additional Claude accounts detected on the machine (beyond the default `~/.claude` login). Exposed
+    /// so the Customize detail can rename an extra account. See the Claude multi-account note below.
+    let claudeAccounts: ClaudeAccountsStore
     /// The provider runtimes, kept so on-demand credential detection (the Customize "Reset All" reseed)
     /// can re-probe `hasLocalCredentials()` the same way first-run seeding does.
     private let providers: [ProviderRuntime]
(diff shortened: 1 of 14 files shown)


robinebers:
🤣
image

robinebers:
Thank you so much @QuadDepo 

This will ultimately be superseded by #987 but requires a lot more work from our end.

We took a good bit of this PR as inspo and will co-credit you accordingly!

QuadDepo:
Thanks @robinebers! Happy to help out <3

PR #975, opened 2026-07-12
feat(providers): fold pi coding agent usage into Claude and Codex

### LINKED ISSUE #974: Fold pi coding agent usage into the Claude and Codex cards
## Problem

The [pi coding agent](https://github.com/earendil-works/pi) is BYO-key: it drives other providers' models (Claude, Codex/OpenAI, Cursor, and more) using your existing subscriptions. When I use my Claude sub inside pi, that usage is billed against my Claude plan, but it never appears on OpenUsage's Claude card, because the Claude card only reads Claude Code's own logs under `~/.claude/projects/`. So my Usage Trend and spend tiles undercount my real Claude usage whenever I code through pi. The same gap exists for Codex.

## Proposed Solution

Read pi's local session logs (`~/.pi/agent/sessions/**/*.jsonl`) and attribute their usage back to the provider that was actually used, folding it into that provider's existing Usage Trend and spend tiles. No new card: pi is a driver, so its usage belongs on the underlying provider.

- pi records an authoritative per-message `usage.cost.total` (like OpenCode), so cost uses that when present, else the shared pricing engine, matching the existing log scanners' rule.
- Merge the pi slice into the native scan, so there is no double counting (pi's session ids are distinct from the Claude Code / Codex CLIs).
- Scope this first change to the providers that already have a local usage-trend card: Claude and Codex. Map (but defer) the rest, and note pi providers OpenUsage doesn't track.

pi -> OpenUsage mapping:

| pi `provider` | OpenUsage card | Wired now? |
|---|---|---|
| `anthropic`, `claude-agent-sdk` | Claude | yes |
| `openai-codex` | Codex | yes |
| `cursor` | Cursor | deferred (CSV-based trend) |
| `zai`/`zhipu`, `google-antigravity`, `github-copilot` | Z.ai / Antigravity / Copilot | deferred (no local trend card) |
| `nvidia-nim` | none | future |

## Alternatives Considered

- **A standalone pi card.** Rejected: pi has no subscription of its own, and the usage is really Claude/Codex usage. A separate card would double-count against the provider cards and misrepresent where the spend lands.
- **Route pi lines through the existing Claude/Codex scanners.** Rejected: pi's log shape differs (`usage.input`/`output`, nested `usage.cost.total`), so it needs its own small parser; it then feeds the same shared `DailyUsageAccumulator`/`SpendTileMapper` path.

## Additional Context

I have a working, tested implementation on a fork (full suite green, 12 new tests, verified against real logs) and would open a focused PR (367 lines, well under the size limit) if this is approved. It reuses the shared `IncrementalJSONLScanner` and pricing engine, adds one small `Sources/OpenUsage/Providers/Pi/` module, and touches only `ClaudeProvider`/`CodexProvider` refresh plus a `DailyUsageAccumulator.merged` helper. Happy to adjust scope or naming to fit the project.

jal-co:
## Approved issue

Fixes #974

## TL;DR

Usage that happens inside the [pi coding agent](https://github.com/earendil-works/pi) is billed against the underlying provider's subscription, so this reads pi's session logs and folds that usage into the Claude and Codex cards' Usage Trend and spend tiles.

## What was happening

- pi is BYO-key: it drives Claude, Codex, and other providers' models with your existing subscriptions.
- The Claude/Codex cards only read each tool's own logs (`~/.claude/projects/`, `~/.codex/sessions/`), so any Claude/Codex usage that went through pi was invisible and the tiles/trend undercounted real usage.

## What this changes

- New `Sources/OpenUsage/Providers/Pi/` module:
  - `PiProviderMapping` maps pi's `provider` field to the OpenUsage card, with untracked pi providers noted for future.
  - `PiPaths` resolves `~/.pi/agent/sessions` (honoring `PI_CODING_AGENT_SESSION_DIR` / `PI_CODING_AGENT_DIR`).
  - `PiUsageScanner` is an incremental log scanner reusing the shared `IncrementalJSONLScanner`, shared as one actor so pi logs parse once per refresh.
- Cost uses pi's authoritative per-message `cost.total` when present, else the shared pricing engine, matching the existing log scanners' rule.
- `DailyUsageAccumulator.merged` folds the pi slice into the Claude/Codex native scans before `SpendTileMapper`, with no double counting (pi ids are distinct, and forked-session replays are de-duplicated).
- Only Claude and Codex are wired (both already have a log-based trend). Cursor, Z.ai, Antigravity, and Copilot are mapped but deferred; `nvidia-nim` has no OpenUsage card.
- Docs updated for both providers.

## Heads-up

- Scope is deliberately minimal: no new cards, no layout/`AppContainer`/widget-descriptor changes. Cursor is deferred because its trend is built from a CSV path, not the shared `LogUsageScan` seam.
- The pricing engine is the universal fallback, so extending to Cursor/Z.ai later needs no pricing work (their pi cost is `$0`; the engine prices `composer-2.5`, `glm-5.x`).

## Tests

- Full suite green: 1071 tests, 0 failures; 12 new tests covering parsing, the carried-cost/pricing rule, dedup, the provider mapping, and the scan merge.
- `swift build` clean.
- Verified end-to-end against real logs (a throwaway smoke test, not committed): Codex total matched the reference exactly; Claude matched within the expected day-window bound.

## Screenshots

Layout is unchanged; this only changes the numbers folded into the existing Usage Trend and spend tiles. Below, the Last 30 Days spend tiles and Today/Yesterday now include usage driven through pi.

| Before | After |
|---|---|
| **Claude**<br>Claude before | **Claude**<br>Claude after |
| **Codex**<br>Codex before | **Codex**<br>Codex after |
**Overall Usage**<br>Overall before | **Overall after**<br>Overall After |


robinebers:
This looks great, thank you. Will review this tomorrow!

jal-co:
> This looks great, thank you. Will review this tomorrow!

Sounds good. All ready for ur review.

robinebers (changes requested):
Thanks for this — the direction looks good, and we’d be happy to merge it once one surgical fix is applied.

In `DailyUsageAccumulator.merged`, please preserve unknown cost rather than converting `nil` to `$0`. Skipping entries without a cost is sufficient, since their unknown-model metadata is already carried through. For example:

```swift
guard let cost = model.costUSD else { continue }
```

The separate `now()` calls can technically straddle midnight, but these metrics are intended as an indicative usage estimate rather than accounting-grade data, so I don’t consider that blocking.

jal-co:
> Thanks for this — the direction looks good, and we’d be happy to merge it once one surgical fix is applied.
> 
> In `DailyUsageAccumulator.merged`, please preserve unknown cost rather than converting `nil` to `$0`. Skipping entries without a cost is sufficient, since their unknown-model metadata is already carried through. For example:
> 
> ```swift
> guard let cost = model.costUSD else { continue }
> ```
> 
> The separate `now()` calls can technically straddle midnight, but these metrics are intended as an indicative usage estimate rather than accounting-grade data, so I don’t consider that blocking.

Should be fixed.

PR #977, opened 2026-07-12
Add secure nearby Mac usage sync

robinebers:
## TL;DR

Adds opt-in, same-network usage aggregation between approved Macs. OpenUsage discovers peers with Bonjour, pairs them with a comparison code, encrypts every transfer, and combines only machine-local spend/token history so account-wide quotas are never double-counted.

## What was happening

- OpenUsage could read usage from only the Mac it was running on.
- People using provider CLIs across several computers had no built-in way to combine local token/spend history.
- The loopback API could expose the current Mac's normalized snapshots, but it was intentionally unavailable to the LAN and had no pairing or authentication model.

## What this changes

- Adds a **Nearby Macs** Settings section with an off-by-default **Share Across Macs** toggle, discovery, Connect, matching-code approval, status, and Forget controls.
- Advertises and browses `_openusage._tcp` with Network.framework/Bonjour and adds the macOS Local Network privacy declarations to development and release bundles.
- Uses ephemeral Curve25519 key agreement for pairing/session keys, a six-digit comparison code, per-device secrets stored in Keychain, HMAC mutual authentication, and ChaCha20-Poly1305 payload encryption.
- Exports only normalized Today / Yesterday / Last 30 Days values and Usage Trend points. Credentials, cookies, API keys, raw logs, plans, warnings, and account-wide quota meters stay on their Mac.
- Keeps received snapshots in memory only, removes a peer's contribution when it becomes unavailable, and refreshes available peers after normal and manual provider refreshes.
- Makes dashboard widgets, Total Spend, the menu bar, and the loopback API read the same combined snapshot view.
- Documents setup, aggregation rules, security, privacy, refresh behavior, architecture, and local API behavior.

## Heads-up

- Bonjour browsing requires macOS Local Network permission. The feature does not touch the LAN until the user turns it on.
- The comparison code must match on both Macs before the receiving Mac approves the request; this is what prevents a network intermediary from silently substituting keys during first pairing.
- Only additive machine-local metrics are merged. Session/Weekly/account quotas deliberately remain local because several Macs usually report the same account-wide value.
- The opt-in integration test uses direct listener endpoints after both Bonjour listeners register because some hosted runners block multicast browsing; production always resolves endpoints through `NWBrowser`.
- The local UI automation bridge was unavailable in this environment (`native pipe startup failed`), so the Settings screenshot is still pending for this draft.

## Tests

- `swift test` — 1,067 tests, 4 existing skips, 0 failures
- `OPENUSAGE_RUN_LAN_INTEGRATION=1 swift test --filter LANSyncStoreIntegrationTests` — two real Network.framework listeners, code match + approval, independent secret stores, authenticated encrypted sync in both directions
- `./script/build_and_run.sh verify` — release build, signed bundle staging, launch, and process verification succeeded
- `git diff --check`

## Screenshots

- Pending before ready-for-review: Nearby Macs off, discovered-peer, code-approval, and connected states. The available UI automation runtime failed to start in this environment.

### DIFF
--- a/Sources/OpenUsage/App/AppContainer.swift
+++ b/Sources/OpenUsage/App/AppContainer.swift
@@ -26,6 +26,8 @@ final class AppContainer {
     /// ephemeral secret-code easter-egg state, and the system accessibility flags it yields to. Read by both
     /// the SwiftUI surface and the AppKit panel (`StatusItemController`).
     let transparency: PopoverTransparencyStore
+    /// Opt-in Bonjour discovery and encrypted aggregation of machine-local usage from approved Macs.
+    let lanSync: LANSyncStore
     /// One-time onboarding state (the first-run Customize hint card). Only ever marked pending by
     /// `FirstRunSeeder` on a fresh install, so existing installs never see the card.
     let onboarding: OnboardingStore
@@ -185,16 +187,30 @@ final class AppContainer {
         dataStore.onRefreshOutcome = { [weak telemetry] providerID, outcome, category, manual in
             telemetry?.record(providerID: providerID, outcome: outcome, category: category, manual: manual)
         }
+        let lanSync = LANSyncStore(
+            localSnapshots: { [weak dataStore] in dataStore?.shareableSnapshots() ?? [:] },
+            applyRemoteSnapshots: { [weak dataStore] deviceID, snapshots in
+                dataStore?.setRemoteSnapshots(snapshots, for: deviceID)
+            },
+            removeRemoteSnapshots: { [weak dataStore] deviceID in
+                dataStore?.removeRemoteSnapshots(for: deviceID)
+            }
+        )
+        dataStore.onRefreshCompleted = { [weak lanSync] in
+            await lanSync?.refreshAvailablePeers()
+        }
+        self.lanSync = lanSync
         self.telemetry = telemetry
         self.transparency = PopoverTransparencyStore()
         self.localAPI = LocalUsageServer(state: { [layout, enablement, dataStore] in
             LocalUsageAPI.State(
                 enabledOrderedIDs: layout.providerOrder.filter { enablement.isEnabled($0) },
                 knownIDs: Set(registry.providers.map(\.id)),
-                snapshots: dataStore.snapshots
+                snapshots: dataStore.displaySnapshots
             )
         })
         self.refreshTask = Self.startPeriodicRefresh(dataStore: dataStore, telemetry: telemetry)
+        lanSync.start()
         localAPI.start()
         // Become the notification-center delegate so banners show while frontmost — a menu-bar accessory
         // effectively always is. Notification authorization is requested the first time a trigger is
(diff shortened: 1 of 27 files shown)


robinebers:
## Closing — we're taking a different approach

We pulled this down, tested it live across two Macs, and are closing it in favor of a redesign. Hand-off notes below.

### What we validated

The product concept works and is worth building: discovery, pairing with a comparison code, and combined spend across Macs all functioned after fixes, and the Settings UX (approval card, paired-device rows) felt right. The privacy boundary (never share credentials/quotas, only additive usage) is the correct instinct.

### Bugs found and fixed during live testing (kept on this branch for reference)

- **Discovery never worked**: browsing used `.bonjour`, which delivers results without TXT metadata, so the id/version filter dropped every discovered Mac. Requires `.bonjourWithTXTRecord` (`7bea8aeb`). Went unnoticed because the integration test bypasses Bonjour discovery entirely.
- **Cursor spend double-counted**: Cursor's spend tiles come from the account-wide server CSV export, not machine-local logs — both Macs reported the same dollars and the aggregator added them (`943dc218`). This is the canary for a structural problem: "machine-local vs account-wide" is a per-provider judgment encoded in flags, and every new provider is a chance to silently double-count.
- **Popover UX**: system permission dialogs and pairing clicks dismissed the panel mid-flow; the stub SwiftUI Settings scene opened as a bare empty window (`7bea8aeb`).
- **Connections hung into generic timeouts**: `.waiting`/`.failed` NWConnection states were never handled, so stale endpoints surfaced as "didn't respond in time" (`943dc218`).
- **Forget was one-sided** and the peer showed a cryptic error afterwards; a pairing Mac was listed twice in Settings (`da65736c`).

### Why we're closing rather than iterating

1. **No convergence.** Each Mac renders "my data + whichever peers I could reach at my last refresh." Two Macs routinely show different totals — by design, not by bug. There is no shared truth to converge on.
2. **It merges rendered UI rows, not data.** Combining `MetricLine`s means matching display labels, aligning chart points by index, and patching value strings — inherently fragile.
3. **Heavy machinery for the job.** Bonjour + hand-rolled pairing protocol + key agreement + Keychain secrets + HMAC + framing is ~1,000 lines of security-sensitive code to add up a few dozen daily numbers between Macs the same person owns.

### What we'll do instead

File-based sync over iCloud Drive: each Mac writes a small JSON of raw machine-local daily usage series (provider → per-day cost/tokens) to a shared folder; every Mac reads all files and computes tiles from the union. No discovery, no pairing, no crypto; converges to identical numbers everywhere; works when Macs aren't awake simultaneously; ~10× less code.

### Coordination with #982

#982 restructures the package (shared `OpenUsage` module, app/CLI entry points, `ProviderCatalog`) and introduces a stable machine-facing limits contract with per-descriptor export annotations — and it touches most of the files this branch touches. Plan:

- **Land #982 first**; the sync feature builds on top of the shared module rather than racing it.
- #982 also sets the pattern the new sync should follow: expose **normalized data with explicit per-descriptor semantics**, not UI rows. The sync feature's "machine-local vs account-wide" flag belongs beside #982's `exportingLimit` annotations so both contracts read from one source of truth.
- #982's limits contract is point-in-time and account-wide, so it does not replace cross-Mac aggregation (which needs machine-local *daily history*) — the two are complementary, and a future `openusage.history.v1` shape could serve both the CLI and the iCloud sync files.

PR #979, opened 2026-07-12
Fix Codex weekly-only window labeling

### LINKED ISSUE #978: Codex weekly quota shows as "session" and 5h quota does not appear
### Description

After the update to 0.7.4, the Session bar shows the weekly amount, not the 5 hour amount, and the Weekly bar does not show anything.

### Steps to Reproduce

1. Open OpenUsage
2. Expand window to show details
3.  Notice that there is only one colored bar for Codex under "Session".
4. Compare to Usage Remaining in Codex Desktop to see it is the Weekly usage remaining, not the Session usage remaining

### Expected Behavior

It should show two bars, one with the correct Session usage remaining for the 5h session, and under it the Weekly bar should show how much of the weekly quota remains.

### Actual Behavior

One bar showing the wrong data.

### OpenUsage Version

0.74

### macOS Version

26.5.2

### Mac Type

Apple Silicon

### Screenshots

Image

Image

Image

### Additional Context

Thank you for OpenUsage it is incredibly helpful!

xuing:
## TL;DR

Identify Codex rate-limit windows by their reported duration instead of assuming `primary_window` is always the five-hour Session limit. A weekly-only response now shows under Weekly, and the five-hour row comes back automatically when Codex restores that limit.

Fixes #978.

## What was happening

- Codex temporarily removed the five-hour usage restriction for Plus, Business, and Pro plans ([announcement](https://x.com/thsottiaux/status/2076365965915467978)).
- With the five-hour window absent, the remaining seven-day limit ships in `primary_window`.
- The mapper labeled `primary_window` as Session unconditionally, so the weekly percentage showed under Session and the Weekly row read "No data".
- The `x-codex-primary-used-percent` header fallback could repeat the same weekly window as a second Session meter.

## What this changes

- Classifies each window as Session or Weekly by its `limit_window_seconds` (5 h vs. 7 days), keeping the usual primary/secondary slot as the fallback when a response omits the duration or reports an unrecognized one.
- Routes the header fallback through the same classification, so a weekly-only primary window can't reappear as Session; a window whose label is already mapped is skipped.
- Applies the same resolution to the model-specific Spark / Spark Weekly windows, which reuse the identical shape.
- Refactors the six near-identical window-mapping blocks into one `appendWindowLine` helper (fresh-window 1%→0 normalization and reset handling unchanged).
- Updates `docs/providers/codex.md` to describe the duration-based labeling.

## Heads-up

- Nothing hardcodes the temporary removal: when Codex restores the 5-hour window (`limit_window_seconds: 18000`), Session and Weekly both map normally with no further change.
- In the degenerate case where both body windows report the same duration, only the first is kept — previously the second would have been shown under a wrong label.

## Tests

- New regression tests: weekly-only `primary_window` (with a conflicting primary header) maps to Weekly with no Session row, and the restored 5h + weekly shape maps both rows.
- A weekly-only Spark `primary_window` likewise maps to Spark Weekly with no Spark row.
- `swift test --filter CodexUsageMapperTests`: all 22 tests pass.
- Full `swift test`: the only failures are 4 pre-existing ones on clean `main` in local token-spend pricing tests (environment-dependent), unrelated to this change.

### DIFF
--- a/Sources/OpenUsage/Providers/Codex/CodexUsageMapper.swift
+++ b/Sources/OpenUsage/Providers/Codex/CodexUsageMapper.swift
@@ -8,6 +8,12 @@ struct CodexMappedUsage: Equatable, Sendable {
 enum CodexUsageMapper {
     static let sessionPeriodMs = MetricPeriod.sessionMs
     static let weeklyPeriodMs = MetricPeriod.weekMs
+
+    private enum WindowKind {
+        case session
+        case weekly
+    }
+
     /// Codex flex credits are worth 4¢ each; the credits line leads with the dollar value
     /// (mirrors the JS plugin's `CREDIT_USD_RATE`).
     static let creditUSDRate = 0.04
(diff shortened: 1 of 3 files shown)


robinebers:
Thank you for the fix — really appreciate you jumping on #978 so quickly!

#980 was merged instead (same duration-based window routing). Closing this one in favor of that.

PR #994, opened 2026-07-13
fix: recover the status item when the menu bar parks it under the notch

### LINKED ISSUE #993: Status item parked under the MacBook notch: app becomes invisible and unreachable (logs say "Status item ready")
### What happens

On a notched MacBook with a crowded menu bar, macOS parks the OpenUsage status item **underneath the notch**. The app keeps running and the log happily reports `[statusitem] Status item ready (button: true, …)`, but there is nothing to see or click anywhere — to the user, OpenUsage simply never launched. Relaunching doesn't help; the item lands in the same spot every time.

**Before** (item exists at x 767–937, notch spans 791–1012 — the strip is invisible, the gap in the middle is the notch):

before

### Environment

- OpenUsage 0.7.4 (reproduced on `main` dev builds too)
- macOS 26.5.2 (25F84), MacBook Pro 14" (logical 1800×1169, notch x 791–1012 per `auxiliaryTopLeftArea`/`auxiliaryTopRightArea`)
- No menu bar manager (Bartender/Ice) involved

### Root cause (measured, not guessed)

macOS lays status items out right-to-left; the item's **right edge is pinned by its neighbors**. Once the right-of-notch area fills up, the leftmost third-party item overflows under the notch. Measured via the Accessibility API:

| Icon style | Measured frame | Visible |
|---|---|---|
| Text (170 pt) | x 767–937 | 24 pt sliver |
| Bars (36 pt) | x 902–938 | **0 pt** |

Two dead ends worth recording:

- **Shrinking the item can never fix it**: the right edge (≈937, inside the notch) is dictated by the neighbors, so narrowing only moves the *left* edge rightward — the measured bars style (36 pt) is *more* hidden than the text strip.
- There is **no public API to reposition** an `NSStatusItem`.

### Proposed fix (implemented, works — PR ready)

Three layers, all driven by one tested pure-geometry check (`auxiliaryTop*Area` vs the button's screen rect, occluded when ≥75% hidden):

1. **Self-rescue (primary)** — rewrite the item's remembered position (the same `NSStatusItem Preferred Position` default that ⌘-dragging maintains) so it re-lands just right of the notch, and bounce `autosaveName` so AppKit re-reads it. Verified live on macOS 26.5: a 99%-hidden item moved back into the visible menu bar.

   rescued

2. **Surrogate button (fallback)** — the preferred-position key is undocumented, so if a future macOS stops honoring it (verified by re-measuring after the rescue), a small round OpenUsage button appears just below the menu bar next to the notch: click opens the dashboard, drag to move, right-click → Hide Until Relaunch.

   pill

3. **Panel anchoring** — whenever the panel is opened while the item is hidden (global shortcut, surrogate), it anchors at the nearest visible notch edge instead of materializing centered under the notch.

Heads-up on the trade-offs: the rescue is zero-sum (a neighbor gets nudged left — deliberate, since the user is trying to reach OpenUsage), and it leans on an undocumented-but-long-stable defaults key, which is exactly why layer 2 exists.

I have this implemented on a branch with unit tests (the geometry fixtures are the real measurements above) and screenshots; happy to open the PR once this issue is approved. Related previously-closed reports that look like the same underlying cause: #202, #517, #573.

maxmilian:
### TL;DR

When a crowded menu bar pushes the status item under the MacBook notch, OpenUsage becomes invisible and unclickable while its log reports "Status item ready". This PR detects that state and recovers: it moves the item back into the visible menu bar, falls back to a small surrogate button if the move doesn't stick, and anchors the panel beside the notch instead of under it.

Closes #993 (opened first per the issue-first flow; measurements and dead-end analysis live there).

### What was happening

- macOS lays status items right-to-left; when the right-of-notch area fills up, the leftmost third-party item overflows under the notch (measured: text strip at x 767–937, bars glyph at x 902–938, notch 791–1012 — zero to 24 pt visible).
- The app keeps running and logs `Status item ready (button: true, …)`, so to the user OpenUsage "never launched", and relaunching lands it in the same spot.
- Shrinking the item cannot help: its right edge is pinned by its neighbors, so narrower styles are *more* hidden. There is no public API to reposition an `NSStatusItem`.
- Opening the panel (e.g. via the global shortcut) anchored it below the hidden button — a window materializing centered under the notch.

### What this changes

- **`NotchGeometry` (new, pure)** — computes the notch rect from `auxiliaryTop*Area`, how much of the button it hides (occluded at ≥75%), and the fallback panel anchor. Mirrors the `PanelGeometry` pattern; unit-tested with the real measured fixtures.
- **Self-rescue (primary)** — `StatusItemController.repositionStatusItemRightOfNotch()` rewrites the item's remembered position (the same `NSStatusItem Preferred Position` default that ⌘-dragging maintains) and bounces `autosaveName` so AppKit re-reads it; the item re-lands just right of the notch. Verified live on macOS 26.5.2: a 99%-hidden item moved to the visible menu bar.
- **`StatusItemOcclusionMonitor` (new)** — checks after launch, on screen-parameter changes, and on a slow 30 s poll. One rescue per occlusion episode; it re-measures afterwards and only if the item is still hidden shows a **surrogate button** (borderless non-activating panel, icon-only) just below the menu bar at the nearest visible notch edge — click opens the dashboard, drag to move, right-click → Hide Until Relaunch. Hidden while the dashboard is open, and removed automatically the moment the real item is visible again.
- **Panel anchoring** — `showPanel()` anchors at the nearest visible notch edge whenever the button is effectively hidden, so the panel never opens centered under the notch.
- **Docs** — `docs/menu-bar.md` gains a short "The notch" section.

### Heads-up

- The rescue leans on the undocumented-but-long-stable preferred-position defaults key (it is how every item's position persists). If a future macOS stops honoring it, the post-rescue re-measure catches that and the surrogate button takes over — behavior degrades to "still reachable", never back to "invisible".
- The rescue is zero-sum: re-landing right of the notch nudges a neighbor leftward. Deliberate — the user is trying to reach OpenUsage.
- Non-notched Macs: `auxiliaryTop*Area` is `nil`, every new path is inert, zero behavior change.
- An earlier IUO footgun is documented in-code: interpolating `statusItem.autosaveName` directly writes a key literally named `… Optional("Item-0")` that AppKit never reads.

### Tests

- `NotchGeometryTests` (8 cases) — fixtures are the real AX measurements from the reproducing machine (screen 1800×1169, notch 791–1012, item frames 767×170 / 902×36).
- Full suite: the only failures are the 4 Codex tests that already fail on a clean `main` checkout.
- Live verification on macOS 26.5.2 (notched 14"): rescue moves the item into view (screenshot below); with the rescue defeated, the surrogate button appears and opens the panel anchored at the notch edge (`anchoring the panel at x=1012` in the log).

### Screenshots

Status item rescued back into the visible menu bar (Claude/Codex/Z.ai strip, right of the notch):

rescued

Surrogate button (fallback path) below the menu bar at the notch's right edge:

pill

https://claude.ai/code/session_019dpF8491eHK8reiuJ8vMxB

### DIFF
--- a/Sources/OpenUsage/App/StatusItemController.swift
+++ b/Sources/OpenUsage/App/StatusItemController.swift
@@ -39,6 +39,19 @@ final class StatusItemController: NSObject {
         onInsidePanelClick: { [weak self] in self?.clearStrayFocus() },
         onDismiss: { [weak self] in self?.hidePanel() }
     )
+    /// Keeps the app reachable when the menu bar overflows and macOS parks the status item under
+    /// the notch: first tries to move the item back into the visible menu bar, and if that doesn't
+    /// stick, shows a surrogate pill at the nearest visible notch edge that opens the dashboard.
+    private lazy var occlusionMonitor = StatusItemOcclusionMonitor(
+        statusItem: statusItem,
+        onActivate: { [weak self] in
+            self?.container.layout.screen = .dashboard
+            self?.showPopover()
+        },
+        onRescue: { [weak self] _ in
+            self?.repositionStatusItemRightOfNotch()
+        }
+    )
     private let hostingController: NSHostingController<AnyView>
     /// The panel's backdrop: an opaque tray by default, swapped to a behind-window vibrancy view when
     /// the transparency style is non-opaque. Built once and toggled, so it can't race the style observer.
@@ -120,6 +133,7 @@ final class StatusItemController: NSObject {
         }
 
         heightController.installBridge()
+        occlusionMonitor.start()
 
         AppLog.info(.statusItem, "Status item ready (button: \(self.statusItem.button != nil), shortcut: \(KeyboardShortcuts.getShortcut(for: .togglePopover)?.description ?? "none"))")
     }
@@ -188,6 +202,50 @@ final class StatusItemController: NSObject {
         // Left-click toggles the popover; right-click (or control-click) drops the context menu.
         // Both arrive through `statusButtonClicked`, which branches on the triggering event.
         button.sendAction(on: [.leftMouseUp, .rightMouseUp])
+    }
+
+    /// Moves a notch-hidden status item back into the visible menu bar by rewriting its preferred
+    /// position — the same defaults key macOS maintains when the user ⌘-drags the item — and
+    /// re-adding the item so AppKit reads it. The key is undocumented but long-stable (it is how
+    /// every status item's position persists across launches); if a macOS release stops honoring
+    /// it, the occlusion monitor's re-measure simply falls back to the surrogate pill.
+    ///
+    /// The menu bar is zero-sum: landing right of the notch shifts a neighbor leftward. That is the
+    /// deliberate trade — the user is interacting with OpenUsage, and the displaced neighbor is
+    /// whatever macOS was about to hide anyway when space next runs out.
+    private func repositionStatusItemRightOfNotch() {
+        guard
+            let button = statusItem.button,
+            let window = button.window,
+            let screen = window.screen,
+            let notch = NotchGeometry.notchRect(of: screen)
+        else { return }
+        let itemWidth = max(button.bounds.width, 24)
+        // Preferred position ≈ distance from the screen's right edge to the item's right edge.
+        // Target: the item's left edge lands just right of the notch.
+        let offset = screen.frame.maxX - notch.maxX - itemWidth - 12
+        guard offset > 0 else { return }
+        // `autosaveName` is an IUO; interpolating it directly would write a key literally named
+        // "… Optional(\"Item-0\")" that AppKit never reads.
+        guard let name = statusItem.autosaveName else { return }
+        UserDefaults.standard.set(
+            Double(offset),
+            forKey: "NSStatusItem Preferred Position \(name)"
+        )
+        // AppKit only reads a preferred position when an autosave name is assigned (toggling
+        // `isVisible` re-adds the item at its old spot). Bounce through a scratch name and back:
+        // assigning the original name again makes AppKit look up — and move to — the position
+        // just written. The scratch key is removed so it can't shadow anything later.
+        statusItem.autosaveName = "\(name).rescue"
+        statusItem.autosaveName = name
+        UserDefaults.standard.removeObject(forKey: "NSStatusItem Preferred Position \(name).rescue")
+        // The bounce can rebuild the button; re-attach click handling and repaint the strip.
+        configureStatusItem()
+        imageUpdater.update()
+        AppLog.warn(
+            .statusItem,
+            "Status item was behind the notch; repositioned to \(Int(offset))pt from the right edge"
+        )
     }
 
     // MARK: - Transparency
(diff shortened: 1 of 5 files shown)


maxmilian:
Still relevant — flagging that rather than letting the timer run out.

The state here is a label, not a technical blocker. @robinebers reviewed the approach on #993 on Jul 18 (*"Gating it to 26 and below would be acceptable if the UX gets this right yes. Will keep this open and take a look."*), and I pushed that version gate on Jul 20 (`a17ea1a`): `NotchGeometry.fallbackIsNeeded(onMacOSMajorVersion:)` returns `false` for 27+, so the whole fallback — occlusion monitor, pill, panel anchoring — is inert on Golden Gate and can be deleted in one sweep whenever macOS 26 support is dropped. `swift build` is clean and `NotchGeometryTests` is green; the branch is rebased and `MERGEABLE`.

It's still a draft only because of the issue-first gate: #993 hasn't been marked `approved` yet. Happy to flip it to ready the moment that lands — and equally happy to close it if the Golden Gate timing makes it not worth carrying. Just don't want the bot to make that call by default.

maxmilian:
Still relevant — but this is the second time I'm answering the bot on this PR, so rather than just saying that again, here is where it actually stands and a concrete way to end the loop.

**The fix is still the only one for this bug.** Re-checked against `main` today:

- `Sources/OpenUsage/Support/NotchGeometry.swift` does not exist on `main` — nothing upstream has taken this on.
- The only menu-bar commit since this PR opened is `Hide menu bar usage while the screen is shared (#1010) (#1013)` (Jul 16), which is orthogonal to notch occlusion.
- #993 is still open, and the reproduction in it (AX coordinates, occlusion-ratio table) is unchanged.

**Everything you asked for is in.** On Jul 18 you said *"Gating it to 26 and below would be acceptable if the UX gets this right yes. Will keep this open and take a look."* — `a17ea1a` (Jul 19) does exactly that: `NotchGeometry.fallbackIsNeeded` gates on `majorVersion <= 26`, so on 27+ the monitor never starts and the panel-anchor branch is skipped. It has been sitting on that commit since.

**What's actually blocking is a label, not code.** The PR carries `needs-approved-issue`, and `stale.yml` exempts only `pinned,security,keep-open,approved,gate-passed` — none of which #993 or this PR has. So the 7-day timer keeps restarting and I keep answering it, which is noise for both of us.

@robinebers, whichever of these you prefer is fine by me:

1. **You want it** → marking #993 `approved` (or this PR `keep-open`) stops the timer and I'll take it out of draft. It's `BEHIND` `main` right now; say the word and I'll rebase — I've held off rather than pushing commits just to keep a bot quiet.
2. **You want changes first** → tell me what the UX still gets wrong and I'll push; the 495 lines here include 8 measured fixtures in `NotchGeometryTests.swift`, so it's cheap for me to adjust.
3. **You don't want it** → say so and **I'll close it myself today**. A `wontfix` from you is a perfectly good outcome; I'd just rather it be your call than an inactivity timer's.

No rush on my end, and no need to reply to this if option 1 is just a label click.

maxmilian:
Closing this myself.

No answer came back on the three options above, and I'd rather take it off your queue than keep resetting a 7-day timer on a PR you haven't had a chance to weigh in on. To be clear about what this is and isn't: you never said you didn't want it — I'm not reading silence as a decision, I'm just not leaving it for the bot to make.

The work isn't lost. The branch stays up, `a17ea1a` still has the version gate you asked for on Jul 18 (`fallbackIsNeeded` returns `false` for macOS 27+, so the whole fallback path is inert on Golden Gate), and `NotchGeometryTests` still has the 8 measured fixtures. If #993 ever gets `approved`, say the word and I'll reopen and rebase — that's a couple of minutes of work, not a restart.

Thanks for the Jul 18 steer on gating; it was the right call and it made the diff considerably smaller than what I first proposed.

PR #996, opened 2026-07-13
feat(cursor): spend view toggle for included vs billed usage

### LINKED ISSUE #609: Cursor: toggle between included usage and raw API usage
## Problem

Cursor tracks two different kinds of usage:

| Kind | Meaning |
|---|---|
| **Included usage** | Consumption covered by the plan (CSV `Cost` column = `Included`) |
| **Raw API usage** | Billed on-demand / API usage (CSV `Cost` column is a dollar amount) |

OpenUsage currently **mixes both** when showing Cursor spend. The Today / Yesterday / Last 30 Days tiles price every CSV row from token counts via the bundled model manifest, ignoring whether Cursor marked the row as included or billed.

That makes the spend tiles misleading for users who only care about one view — e.g. "how much of my included plan did I use today?" vs "how much API/on-demand spend did I rack up?"

## Current behavior

- `CursorUsageCSV.parse` reads token columns but **drops the `Cost` column** (see comment in `CursorUsageCSV.swift`: actual-cost / CostMode path deferred for v1).
- `CursorUsageMapper.appendSpendLines` sums `imputedCostDollars` across all rows with no filter.
- Dashboard API metrics (`apiPercentUsed`, `autoPercentUsed`, `totalPercentUsed`) come from Cursor's Connect RPC and are separate from the CSV spend tiles.

## Ask

Add a user-facing toggle (Cursor-specific) to choose which usage view to show for spend-related metrics:

- **Included only** — aggregate rows where `Cost` = `Included` (priced locally from tokens, or $0 if we treat included as non-billable)
- **Raw API only** — aggregate rows with an actual dollar `Cost` from the CSV (or non-`Included` rows)
- **All usage** (default) — current behavior for users who want the combined picture

### UX ideas

- Setting in **Settings** under a Cursor section, or a compact control on the Cursor provider card
- Clear labels: e.g. "Included Usage" / "API Usage" / "All Usage"
- Persist preference per device
- Spend tiles (Today, Yesterday, Last 30 Days) respect the toggle; document whether bounded metrics (Total Usage %, API Usage %) should also switch or stay as Cursor reports them

## Implementation notes

- Extend `CursorUsageCSVRow` to carry billing kind (`included` vs `billed`) parsed from the `Cost` column.
- Thread the preference through `CursorProvider` → `appendSpendLines` (and any Tauri plugin parity if still maintained on `main`).
- Add tests for mixed CSV rows (some `Included`, some `$1.23`) and verify each toggle mode sums correctly.
- Update `docs/providers/cursor.md` to explain the toggle and what each mode means.

## References

- `Sources/OpenUsage/Providers/Cursor/CursorUsageCSV.swift`
- `Sources/OpenUsage/Providers/Cursor/CursorUsageMapper.swift` — `appendSpendLines`
- `Sources/OpenUsage/Providers/Cursor/CursorProvider.swift` — spend widget descriptors
- Prior art: cursorcat `UsageCSV.swift` CostMode path (dropped in v1 port)
- Test fixtures in `Tests/OpenUsageTests/CursorSpendTests.swift` already use `Cost` = `Included`

## Acceptance

- User can switch between included-only, raw-API-only, and all-usage views for Cursor spend tiles.
- Each mode produces correct totals on mixed CSV data.
- Default preserves today's combined behavior.
- Documented in `docs/providers/cursor.md`.

maxmilian:
### TL;DR

Adds a Cursor **Spend View** toggle (All / Included / API Usage) so the spend tiles and Usage Trend can show plan-included usage, billed on-demand usage, or the combined picture — implementing #609.

Closes #609.

### What was happening

- Cursor's export marks each row as plan-**Included** or billed (a dollar `Cost`), but OpenUsage dropped the `Cost` column and summed every row together.
- The Today / Yesterday / Last 30 Days tiles and the Usage Trend therefore always mixed both kinds, so a user couldn't answer "how much of my included plan did I use?" vs "how much on-demand spend did I rack up?"

### What this changes

- **Parse billing kind** — `CursorUsageCSV` reads the `Cost` column into a new `CursorBillingKind` (`included` / `billed` / `unknown`). The column is optional: exports without it still parse, with rows left `.unknown`.
- **Spend View setting** — new `CursorSpendViewSetting` (`all` / `included` / `api`), a `UserDefaultsBacked` enum like `TimeFormatSetting`, persisted per device, default `all` (unchanged behavior).
- **Filtering** — `CursorUsageMapper.appendSpendLines` takes the view and filters rows before aggregating; a view with no matching rows reads "No data" rather than a fabricated \$0. Filtered rows drop out of tiles, trend, and the model breakdown alike.
- **Provider** — `CursorProvider` reads the setting at each refresh (so a flip applies on the next snapshot) and logs when a Cost-less export can't satisfy a non-All view.
- **UI** — a Cursor-only **Spend View** card in the Cursor Customize detail (`CursorSpendViewSection`), following the `APIKeysSection` pattern; changing it forces a refresh.
- **Docs** — `docs/providers/cursor.md` gains a Spend View subsection.

### Heads-up

- Per the issue's open question: the bounded meters (Total / Auto / API Usage %) intentionally **stay as Cursor reports them** — they come from Cursor's live API, not the export, and re-deriving them locally would contradict the dashboard. The toggle scopes to the CSV-derived spend tiles + trend only.
- `unknown` rows (Cost-less exports) count only under All Usage; assigning them to a side would fake the split.

### Tests

- `CursorSpendViewSetting` filtering: All = \$7 / 700 tok, Included = \$1 / 100 tok, API = \$2 / 200 tok on a mixed day; empty-after-filter → "No data".
- `Cost` column parsing: `Included`/`included` → included, `\$1.23`/`1.23` → billed, empty/absent column → unknown.
- Full suite green apart from the 4 Codex tests already red on a clean `main`.

### Screenshots

Cursor Customize detail → new **Spend View** card with the three-way picker:

spend view picker

https://claude.ai/code/session_019dpF8491eHK8reiuJ8vMxB

### DIFF
--- a/Sources/OpenUsage/Providers/Cursor/CursorProvider.swift
+++ b/Sources/OpenUsage/Providers/Cursor/CursorProvider.swift
@@ -16,17 +16,21 @@ final class CursorProvider: ProviderRuntime {
     let usageClient: CursorUsageClient
     let now: @Sendable () -> Date
     let pricing: @Sendable () async -> ModelPricing
+    /// Read at each refresh so a toggle flip applies on the next snapshot without a restart.
+    let spendView: @Sendable () -> CursorSpendViewSetting
 
     init(
         authStore: CursorAuthStore = CursorAuthStore(),
         usageClient: CursorUsageClient = CursorUsageClient(),
         now: @escaping @Sendable () -> Date = Date.init,
-        pricing: @escaping @Sendable () async -> ModelPricing = { await ModelPricingStore.shared.current() }
+        pricing: @escaping @Sendable () async -> ModelPricing = { await ModelPricingStore.shared.current() },
+        spendView: @escaping @Sendable () -> CursorSpendViewSetting = { .current }
     ) {
         self.authStore = authStore
         self.usageClient = usageClient
         self.now = now
         self.pricing = pricing
+        self.spendView = spendView
     }
 
     var widgetDescriptors: [WidgetDescriptor] {
@@ -186,7 +190,15 @@ final class CursorProvider: ProviderRuntime {
                     "usage CSV ignored \(parsed.rejectedRowCount) malformed row\(parsed.rejectedRowCount == 1 ? "" : "s")"
                 )
             }
-            return CursorUsageMapper.appendSpendLines(rows: parsed.rows, now: end, pricing: pricing, to: &lines)
+            let view = spendView()
+            if view != .all, !parsed.rows.isEmpty, parsed.rows.allSatisfy({ $0.billing == .unknown }) {
+                // An export without a Cost column can't be split; say so instead of a silent "No data".
+                AppLog.warn(
+                    LogTag.plugin("cursor"),
+                    "usage CSV carries no Cost column; the \(view.label) spend view has nothing to show"
+                )
+            }
+            return CursorUsageMapper.appendSpendLines(rows: parsed.rows, now: end, pricing: pricing, view: view, to: &lines)
         } catch let error as CursorUsageCSVError {
             switch error {
             case .missingColumns(let columns):
--- a/Sources/OpenUsage/Providers/Cursor/CursorProvider.swift
+++ b/Sources/OpenUsage/Providers/Cursor/CursorProvider.swift
@@ -16,17 +16,21 @@ final class CursorProvider: ProviderRuntime {
     let usageClient: CursorUsageClient
     let now: @Sendable () -> Date
     let pricing: @Sendable () async -> ModelPricing
+    /// Read at each refresh so a toggle flip applies on the next snapshot without a restart.
+    let spendView: @Sendable () -> CursorSpendViewSetting
 
     init(
         authStore: CursorAuthStore = CursorAuthStore(),
         usageClient: CursorUsageClient = CursorUsageClient(),
         now: @escaping @Sendable () -> Date = Date.init,
-        pricing: @escaping @Sendable () async -> ModelPricing = { await ModelPricingStore.shared.current() }
+        pricing: @escaping @Sendable () async -> ModelPricing = { await ModelPricingStore.shared.current() },
+        spendView: @escaping @Sendable () -> CursorSpendViewSetting = { .current }
     ) {
         self.authStore = authStore
         self.usageClient = usageClient
         self.now = now
         self.pricing = pricing
+        self.spendView = spendView
     }
 
     var widgetDescriptors: [WidgetDescriptor] {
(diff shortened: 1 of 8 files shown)


robinebers:
Nice work — architecture and default look right (`All Usage` / `.all` preserves today's combined spend tiles; meters stay API-sourced; Cost-less rows only count under All).

**Follow-up I'd take:**

1. **Fix the label wrap** — `Spend Tiles Show` breaks onto two lines even with empty space before the picker. SwiftUI is letting the `Text` wrap so the `Spacer` can expand. Pin it with `.lineLimit(1)` (or `.fixedSize(horizontal: true, vertical: false)`).
2. **Shorten the row copy** — the section is already titled Spend View, so `Spend Tiles Show` is redundant. Something like `Show` (or reuse Settings' plain row+picker shape) would read cleaner and reduce wrap pressure.
3. **Optional naming polish** — picker option `API Usage` collides with Cursor's existing API Usage meter. Worth a later pass if that confuses people in practice (`On-Demand` / `Billed`?).

Not blocking on (3); (1)–(2) are the ones I'd land before or right after merge.

maxmilian:
Thanks for the review! Pushed `4ac3e81` addressing (1) and (2) together:

- Row label is now just `Show` — the `Spend View` header already names the card, so the longer copy was redundant.
- Pinned it with `.fixedSize(horizontal: true, vertical: false)` so the trailing `Spacer` can't squeeze the `Text` into a wrap.

On (3): agreed that `API Usage` colliding with Cursor's own API Usage meter is worth clearing up — I'd lean `Billed` (it's the on-demand/over-plan spend). Happy to fold that rename in here if you'd like it now, or leave it for a follow-up per your call since it touches the option labels users will see.

maxmilian:
Folded (3) in and rebased onto current `main` — 23 commits behind; the only two that touched these files are the Account-first passes over `CustomizeProviderDetailView`, which don't overlap the spend view, so the rebase was clean.

On (3) specifically: I asked back on 07-18 whether you wanted the rename now or left as a follow-up. Rather than hold the PR waiting on an answer, I've made the call and kept it trivially reversible — it's its own commit (`53176f5`), so dropping it leaves (1)–(2) intact if you'd rather ship the rename separately or not at all.

`API Usage` → **`Billed`**. I renamed the enum case along with the label (`case api` → `case billed`) instead of only the user-facing string, so the code doesn't carry two words for the same side. The raw value moves with it; nothing to migrate, since this hasn't shipped.

Deliberately left alone: the bounded `API Usage %` meter, and the doc sentence describing Cursor's own export terminology ("plan-**included** vs billed **API** usage"). The collision was the picker borrowing the meter's name — Cursor's term for the billing side isn't itself wrong, so only the option that sat next to the meter needed to move.

`swift build` clean; `swift test --filter CursorSpend` 14/14 green.

PR #1018, opened 2026-07-17
Fix Settings transitions and honor Reduce Animations

davidarny:
## TL;DR

Makes the in-popover Settings screen open promptly and applies Reduce Animations consistently across SwiftUI transitions, AppKit popovers, progress indicators, and time-driven effects.

## What was happening

- Opening Settings waited on a synchronous launch-at-login status lookup, causing a visible delay.
- The dashboard and Settings screen briefly coexisted during navigation, so the new header pushed the existing content down before the transition completed.
- Reduce Animations did not cover several independent motion sources, including hover popovers, indeterminate progress indicators, glass interactions, and timeline-driven effects.
- The separate Usage Trend hosting root did not inherit the dashboard animation-reduction policy.

## What this changes

- Loads launch-at-login status asynchronously and keeps the Settings transition responsive.
- Adds a shared reduction policy that combines the app preference with the macOS accessibility setting.
- Disables page transitions, panel crossfades, time-driven effects, interactive glass motion, symbol effects, and indeterminate spinner clocks when motion is reduced.
- Uses a shared AppKit-backed hover-popover presenter so Usage Trend, model breakdown, and reset-credit popovers can open and close without animation.
- Propagates the reduction policy into the separate popover hosting root.
- Adds regression coverage and documents the setting.

## Heads-up

The reduced-motion hover-popover path uses an AppKit-owned NSPopover because SwiftUI does not expose its animation flag.

## Tests

- `swift test` — 1,216 XCTest tests passed, 3 skipped; 3 Swift Testing tests passed.
- `CONFIG=debug ./script/build_and_run.sh verify` — debug app rebuilt, relaunched, and verified.

## Screenshots

Not applicable: this changes transition timing and motion behavior without changing the static layout.

### DIFF
--- a/Sources/OpenUsage/App/StatusItemController.swift
+++ b/Sources/OpenUsage/App/StatusItemController.swift
@@ -62,6 +62,9 @@ final class StatusItemController: NSObject {
         let hosting = NSHostingController(
             rootView: AnyView(
                 DashboardView()
+                    // Own the preference outside `DashboardView` so the view can read the resolved
+                    // value while deciding whether to mount its two-page transition structure.
+                    .reduceAnimationsWhenRequested()
                     .environment(container)
                     .environment(container.layout)
                     .environment(container.dataStore)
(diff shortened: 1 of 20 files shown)


davidarny:
Superseded by #1019, which presents this work as the new Reduce Animations feature.

PR #1019, opened 2026-07-17
Add Reduce Animations support

davidarny:
## TL;DR

Adds a first-class Reduce Animations preference that works alongside the macOS accessibility setting and consistently suppresses motion throughout the menu-bar popover.

## What was happening

- OpenUsage had no app-level preference for reducing motion.
- SwiftUI transitions, AppKit popovers, progress indicators, glass interactions, symbol effects, and time-driven visuals each managed animation independently.
- Hover popovers required an AppKit bridge because SwiftUI does not expose the underlying presentation animation flag.

## What this changes

- Adds a persisted Reduce Animations setting in General settings.
- Combines the app preference with the macOS Reduce Motion accessibility preference through a shared reduction policy.
- Applies that policy to dashboard and Settings transitions, panel crossfades, time-driven effects, interactive glass motion, symbol effects, and indeterminate progress indicators.
- Adds a shared AppKit-backed presenter for Usage Trend, model breakdown, and reset-credit hover popovers so they can open and close without animation.
- Propagates the reduction policy into separate SwiftUI hosting roots.
- Skips hidden popover layout and sizing while a hover popover is closed.
- Closes AppKit hover popovers synchronously when the menu-bar panel closes.
- Adds regression coverage and documents the new preference.

## Heads-up

The reduced-motion hover-popover path uses an AppKit-owned NSPopover because SwiftUI does not expose its animation flag.

## Tests

- `swift test` — 1,217 XCTest tests passed, 3 skipped; 3 Swift Testing tests passed.
- `swift test --filter UsageTrendPopoverTests` — presentation flags, hidden sizing, and show/dismiss lifecycle passed.
- `CONFIG=debug ./script/build_and_run.sh verify` — debug app rebuilt, relaunched, and verified.
- Manual — with Reduce Animations enabled and a hover popover open, closing via the global shortcut dismissed both windows immediately; reopening did not restore the hover popover.

## Screenshots

Not applicable: the feature changes transition and motion behavior without changing the static layout.


robinebers (changes requested):
Thanks — the feature itself is in good shape. The transaction clamp at the popover root, the structural pager snap, and the pure `MotionTimelineMode` policy all check out, coverage matches what the description claims, and default (setting off) behavior is unchanged everywhere we looked. Requesting changes on three points before merge:

**1. Split the Launch at Login changes into their own PR.** The `LaunchAtLoginSetting` rewrite (async status load + optimistic `isEnabled = enabled` on success) has nothing to do with animations, and it carries the one real regression here: the old code re-read `SMAppService.mainApp.status` after register/unregister, the new code assumes the requested value — so a login item that lands in `.requiresApproval` shows the toggle as ON while the app won't actually launch at login. Happy to review that change on its own merits in a separate PR (with a regression test where `register()` succeeds but the authoritative status stays `false`). The Settings section refactor can stay here.

**2. Gate the hidden layout work in `ReducedMotionPopoverController.update()`.** It runs `host.view.layoutSubtreeIfNeeded()` + `fittingSize` on every `updateNSView`, even while the popover is closed — with Reduce Animations on, every dashboard refresh lays out every row's hidden chart/breakdown detail view. Early-return the sizing when the popover isn't (about to be) presented.

**3. Please verify manually:** with Reduce Animations on and a hover popover open, close the panel via the global shortcut. The AppKit popover's close rides the next SwiftUI update after `HoverPopoverState.dismissAll()`, and if updates defer for the ordered-out panel the popover could linger orphaned on screen.

Minor, non-blocking: on second and later Settings visits the Launch at Login toggle is enabled while a status refresh is in flight, so a stale load can briefly clobber a fresh toggle (moot if #1 moves out); and `UsageTrendPopoverTests` only asserts the static popover flags, not the presentation/dismissal lifecycle the bridge introduces.

davidarny:
@robinebers Addressed all three requested changes in 9fc07fa:

1. Removed the LaunchAtLoginSetting implementation and test changes from this PR; both files now match main.
2. ReducedMotionPopoverController now skips hidden layout and fitting-size work while the popover is not presented, with regression coverage.
3. Panel close now dismisses live AppKit hover popovers synchronously after resetting their SwiftUI state. Added a real show/dismiss lifecycle test, and manually verified the global-shortcut scenario with Reduce Animations enabled: both windows close immediately and the hover popover does not return when reopening.

Full suite: 1,217 XCTest tests passed (3 skipped), plus 3 Swift Testing tests.

lajczi:
hi, does this follow macOS reduce motion setting?

davidarny:
@lajczi yes it does

> Combines the app preference with the macOS Reduce Motion accessibility preference through a shared reduction policy.

PR #1051, opened 2026-07-24
Add Kimi (Moonshot AI) provider

kodin00:
Screenshot 2026-07-25 at 01 06 14

Tracks the Kimi Code subscription's 5-hour session window and 7-day quota from the login the Kimi CLI already leaves on the machine.

Quota comes from `GET https://api.kimi.com/coding/v1/usages`, the endpoint behind the CLI's own `/usage` command. It is not in Moonshot's published API reference, so the mapper degrades to a typed `quotaUnavailable` error rather than a blank card, and docs/providers/kimi.md says the endpoint is unofficial.

Kimi access tokens live only ~15 minutes, so the provider refreshes them itself instead of waiting for the CLI. Because that means writing into a file another tool owns, the rotation re-reads the live credential first (adopting a token the CLI already rotated, rather than replaying a retired one), holds the same flock the CLI's own cross-process refresh takes, and merges into the raw JSON so unmodelled fields survive. A corrupt-but-present file throws instead of being rebuilt from memory.

The flock helper is lifted out of JSONLScanCacheStore into Support/FileLock so both callers share one implementation; the scan cache keeps its 0700 directory assertion via a parameter, and credential locks leave another tool's directory permissions alone.

Metrics: Session (isSessionWindow) and Weekly, both enabled, Always Visible, and pinned — the 2-per-provider cap. No spend tiles: the API reports only percentages, so there is nothing to price.

ResetDisplayTests and ProviderLinksTests hardcode their provider lists and so were passing without covering Kimi; both now include it.

## Approved issue

Fixes #

## TL;DR

Adds Kimi (Moonshot AI) as a provider, tracking the Kimi Code subscription's 5-hour session window
and 7-day quota from the login the Kimi CLI already leaves on the machine. No setup for the user.

## What was happening

- OpenUsage tracked Claude, Codex, Cursor, Copilot, Devin, Grok, OpenCode, OpenRouter and Z.ai, but
  had nothing for Moonshot's Kimi Code subscription.
- The only way to see that quota was typing `/usage` inside the Kimi CLI's interactive REPL — there
  was no way to read it from a dashboard or menu bar. Upstream has an open request for exactly this
  ([kimi-cli#2169](https://github.com/MoonshotAI/kimi-cli/issues/2169)).
- Moonshot's published API reference documents only the pay-as-you-go credit balance, which is a
  different product from the subscription and a different account system.

## What this changes

**New provider module** — `Sources/OpenUsage/Providers/Kimi/` (auth store, usage client, mapper,
provider), registered in `ProviderCatalog` in the alphabetical tail after Grok.

**Two metrics**, both enabled, Always Visible, and pinned (the 2-per-provider cap):

| Metric | Source |
|---|---|
| Session | the 5-hour entry in the response's `limits` array, matched by window duration (`isSessionWindow: true`, so an unused window reads "Not started") |
| Weekly | the response's `usage` block |

No spend tiles — the API reports only percentages, so there is nothing to price and Kimi does not
appear in Total Spend.

**Quota endpoint:** `GET https://api.kimi.com/coding/v1/usages`, the same one behind the CLI's
`/usage` command. Numbers arrive as JSON *strings* (`"100"`, `"41"`), which `ProviderParse.number`
already coerces.

**Credentials:** `$KIMI_CODE_HOME/credentials/kimi-code.json` (default `~/.kimi-code/…`), falling
back to `$KIMI_SHARE_DIR/…` (default `~/.kimi/…`) — the shipped and upstream CLI layouts disagree, so
both are read. Plain 0600 JSON; the CLI retired keyring storage, so there's no Keychain prompt.

**Token refresh.** Kimi access tokens live ~900 seconds, so a read-only borrow would leave the card
blank most of the time. The provider refreshes via `POST https://auth.kimi.com/api/oauth/token` and
writes the rotated pair back into the CLI's own file. Since that file belongs to another tool, the
rotation:

- re-reads the live credential first, adopting a token the CLI already rotated rather than replaying
  a retired one (the `CodexProvider` precedent for `refresh_token_reused`);
- holds an exclusive `flock` on `credentials/kimi-code.lock` — the same lock file the CLI's own
  cross-process refresh takes, so the two genuinely interlock;
- merges into the raw JSON dict so unmodelled fields survive, and **throws** on a
  corrupt-but-present file instead of rebuilding it from memory.

A failed save logs at `.error` and keeps the token for the session only, per the house pattern.

**Shared `flock` helper.** Lifted out of `JSONLScanCacheStore` into `Support/FileLock.swift` so both
callers share one implementation instead of duplicating the boilerplate. The scan cache keeps its
0700 directory assertion via a parameter; credential locks pass nothing, so OpenUsage never
re-permissions a directory another tool owns.

**Plan label** derived from `user.membership.level` by stripping the `LEVEL_` prefix and title-casing
(`LEVEL_INTERMEDIATE` → `Intermediate`). Deliberately *not* mapped onto Moonshot's marketing tier
names (Adagio / Moderato / Allegretto / Allegro / Vivace) — the two vocabularies don't line up
one-to-one, and a wrong guess would label someone's card with a plan they aren't on.

**Also:** `CategorizedError` conformances for both error enums (without them every failure buckets as
`.other`), `moon.stars` symbol fallback, and a `kimi.svg` mark.

## Heads-up

- **The `/usages` endpoint is unofficial.** It is not in Moonshot's published reference; it was found
  in the CLI's own source and verified live. Moonshot can change it without notice, so the mapper
  degrades to a typed `quotaUnavailable` error rather than a blank card or wrong numbers, and
  `docs/providers/kimi.md` says so plainly.
- **Writing into another tool's credentials is the main risk here.** A bug logs the user out of Kimi.
  That's what the flock + re-read-before-refresh + merge-don't-rebuild combination is for, and the
  "`kimi` is still logged in afterwards" check below is the regression that matters most.
- **The icon is a simplified monochrome K-plus-dot mark, not Kimi's official artwork.** Moonshot ships
  only a PNG logo, and `SVGPath` has no arc support, so it's hand-authored from lines and cubics —
  the same approach as the existing `zai.svg`. Happy to swap in an official path.
- **Unmapped rate windows are logged, not dropped.** Moonshot documents a 7-day rate window that
  isn't in the payload today; if it appears it shows up in the log instead of vanishing.
- **Left out of scope:** the Moonshot Open Platform dollar balance (`/v1/users/me/balance`, needs its
  own `sk-` key), `parallel.limit` (a static plan attribute), and the `boosterWallet` extra-usage
  top-up wallet that appears in the payload.

## Tests

`swift test` — **1320 tests, 0 failures.**

New: `KimiProviderTests.swift` (auth store, mapper, provider) and `KimiLayoutTests.swift`. The mapper
suite uses a live-recorded payload verbatim as its happy-path fixture, and covers `remaining`-only
details, a measured zero, scaling against the reported `limit`, an unrecognized rate window, and an
unknown membership level. The provider suite covers pre-emptive refresh, adopting a CLI-rotated
token, 401 → refresh → retry, a second 401 as a hard failure, `invalid_grant`, an unrecognized 400,
and asserts `errorCategory` on every error path.

`ResetDisplayTests` and `ProviderLinksTests` hardcode their provider lists, so they were passing
*without* covering Kimi — both now include it, which is the point of those wiring pins.
`ProviderMarksTests` gains a check that the hand-authored SVG parses into real geometry rather than
silently degrading to the SF Symbol fallback.

Verified live against a real Kimi Code subscription, through the running app:

- auto-detected on first run — `first run: detected credentials for [..., "kimi"]`, proving
  `hasLocalCredentials()`
- plan `Intermediate`; session `used=0` with an 18000s window, weekly `used=41` resetting Jul 31
- waited for the token to enter the 5-minute refresh window and forced a refresh:
  `token refresh ok (rotated)` — access *and* refresh tokens both changed, expiry extended 900s, all
  six credential keys preserved
- **`kimi provider list` still reports `source=oauth` afterwards** — the write-back does not break the
  CLI's session


robinebers:
Hi, we do not approve providers unless they have significant support in the original issue. Cheers

Issue #1054, opened 2026-07-25
Add Kimi (Moonshoot AI) Provider


kodin00:
### Provider Name

Kimi

### Provider Website

https://www.kimi.com

### What usage data does this provider expose?

monthly limit

### Will you build the provider?

Yes, I will submit a PR

urbanscribe:
+1

robinebers:
Duplicate of #791.

Both issues ask for the same thing: bringing the Kimi (Moonshot) provider into the native Swift edition. #791 is the older thread and already carries the full history — the maintainer's requirements for a new provider (a test account made available to the project, plus enough public demand), the closed implementation PR #1028, and the ongoing discussion about whether a console API key is sufficient in place of a dedicated account.

Consolidating there so the demand signal and the open "key vs. account" question stay in one place. Please add any +1s or offers of test credentials to #791.

Note: #1076 (`kimi-k3-high` showing an unknown-model warning under Cursor) is a separate issue — that one is about Cursor's pricing supplement, not about a Kimi provider.

PR #1049, opened 2026-07-24
feat(kimi): add native Kimi Code provider (revival of #1028)

bzqzheng:
**TL;DR** — Adds a native Kimi Code provider (session + weekly quota meters, plan label) by reviving the closed #1028, rebased onto current `main` and re-verified live against Kimi's API today. Refs #791.

**What was happening**

- v0.7's native provider system shipped without Kimi Code; its users lost usage tracking (#791, approved, with steady community demand since).
- #1028 implemented exactly this provider, but was closed over a process concern — no one on the team has a Kimi account to test/maintain with — not over code quality. The contributor has kept the branch current since.
- Kimi Code's own migration (Kimi CLI → Kimi Code CLI) moved the credential file from `~/.kimi/` to `~/.kimi-code/`; any implementation has to read both, new path first.

**What this changes**

- Restores @kotuke's #1028 commits **with authorship preserved** (I only rebased them onto `main` @ `9d2bf09`): a `Providers/Kimi/` module — auth store (Kimi Code CLI OAuth tokens, current + legacy paths, plus optional user API key), usage client (CLI-equivalent token refresh, retry-once-on-401), mapper (Session/Weekly percent meters, tempo plan names), plus registration, icon, `DefaultLayout`, local-API exports, README, and `docs/providers/kimi.md`.
- Includes kotuke's follow-up fix mapping Kimi's internal membership enum to the plan names users actually see (e.g. `LEVEL_INTERMEDIATE` → "Allegretto"), sourced from Kimi's public `ListGoods` catalog.
- Adds a review-hardening commit on top (my own): an independent review found the CLI-owned credential file was treated as fully owned — a strict all-or-nothing decode could misdiagnose a signed-in user as "Not logged in", and `save()` stripped unknown CLI fields. Both fixed (tolerant per-field decode with loud logging, round-trip merge save), plus: near-expiry access tokens without a refresh token are now tried instead of declared dead, sub-second window units can't produce a 1000x period error, wrongly-typed quota containers are invalid-response instead of "No usage data", four stale `docs/` pages synced, and 14 new tests (refresh-400, API-key-401, refresh-2xx-without-token, usage-500, fractional `expires_at`, the live-file re-read race, and more).

**Heads-up**

[...]

Issue #791, opened 2026-06-29
Feature request: restore Kimi provider in the native v0.7 provider system


MumuTW:
Hi, I saw in #565 that Kimi was removed from v0.7 because the team did not have an account to test with.

I have an active Kimi Code account and can help test a native Swift `KimiProvider`. I can provide redacted API response shapes, local verification logs, and feedback against the current v0.7 provider system.

For context, the old JS provider used:

```text
GET https://api.kimi.com/coding/v1/usages
```

The current Kimi Code CLI credential file on macOS is:

```text
~/.kimi-code/credentials/kimi-code.json
```

This is related to the old credential-path reports in #574 and #581, but the request here is different: Kimi is no longer present in the native v0.7 provider registry, so the old plugin cache/path fix does not restore live refresh in v0.7.

Would you accept a native Swift `KimiProvider` contribution with tests? If external PR creation is currently restricted, what is the preferred way to submit the patch: a fork branch, an issue-attached diff, or a maintainer-invited PR?

I will not share tokens or credential contents, only redacted response schemas and verification output.

robinebers:
Hi, thanks for the report.

Makes sense, but unfortunately we don't hold accounts for this. So unless Kimi provides us with keys to maintain this, it would be hard to test and maintain as [mentioned here](https://github.com/robinebers/openusage/issues/783#issuecomment-4830222405) for Z.ai/GLM.

soilSpoon:
> Hi, thanks for the report.
> 
> Makes sense, but unfortunately we don't hold accounts for this. So unless Kimi provides us with keys to maintain this, it would be hard to test and maintain as [mentioned here](https://github.com/robinebers/openusage/issues/783#issuecomment-4830222405) for Z.ai/GLM.

I am currently using Kimi Code, Factory Droid, and Ampcode. Would it be possible for me to submit PRs for these providers to participate in their maintenance?

robinebers:
> > Hi, thanks for the report.
> > Makes sense, but unfortunately we don't hold accounts for this. So unless Kimi provides us with keys to maintain this, it would be hard to test and maintain as [mentioned here](https://github.com/robinebers/openusage/issues/783#issuecomment-4830222405) for Z.ai/GLM.
> 
> I am currently using Kimi Code, Factory Droid, and Ampcode. Would it be possible for me to submit PRs for these providers to participate in their maintenance?

Hi, the way to do this would be to create issues and measure support. All three could come in, but they require:

[...]

PR #1055, opened 2026-07-25
Add Kimi (Moonshot AI) provider

kodin00:
https://github.com/robinebers/openusage/issues/1054

Tracks the Kimi Code subscription's 5-hour session window and 7-day quota from the login the Kimi CLI already leaves on the machine.

Quota comes from `GET https://api.kimi.com/coding/v1/usages`, the endpoint behind the CLI's own `/usage` command. It is not in Moonshot's published API reference, so the mapper degrades to a typed `quotaUnavailable` error rather than a blank card, and docs/providers/kimi.md says the endpoint is unofficial.

Kimi access tokens live only ~15 minutes, so the provider refreshes them itself instead of waiting for the CLI. Because that means writing into a file another tool owns, the rotation re-reads the live credential first (adopting a token the CLI already rotated, rather than replaying a retired one), holds the same flock the CLI's own cross-process refresh takes, and merges into the raw JSON so unmodelled fields survive. A corrupt-but-present file throws instead of being rebuilt from memory.

The flock helper is lifted out of JSONLScanCacheStore into Support/FileLock so both callers share one implementation; the scan cache keeps its 0700 directory assertion via a parameter, and credential locks leave another tool's directory permissions alone.

Metrics: Session (isSessionWindow) and Weekly, both enabled, Always Visible, and pinned — the 2-per-provider cap. No spend tiles: the API reports only percentages, so there is nothing to price.

ResetDisplayTests and ProviderLinksTests hardcode their provider lists and so were passing without covering Kimi; both now include it.

## Approved issue

Fixes #


kotuke:
Heads-up on prior art here, in case it saves you duplicated effort.

This provider has been implemented twice before, both against #791 — which carries the `approved` label, unlike #1054. That may matter for the gatekeeper bot.

- #1028 — my original implementation. Closed by a maintainer over maintenance, not code quality: nobody on the team has a Kimi account to regression-test with.
- #1049 — @bzqzheng's rebase of it onto current `main`, plus a review-hardening pass. Closed voluntarily for the same reason.

Both branches are still available. Comparing them against yours, things **yours** does better and worth keeping:

- the `flock` on the CLI's own `kimi-code.lock` around rotation — ours merges into the file but never locks it, so a concurrent `kimi` refresh can race it;
- `isSessionWindow: true` on the session tile (we missed it entirely);
- the `KIMI_CODE_HOME` / `KIMI_CODE_BASE_URL` overrides;
- treating a 400 as expiry only when the body actually says `invalid_grant`.

Things **ours** has that yours doesn't, which you may want to pull in:

- an API-key path (`~/.config/openusage/kimi.json` + `KIMI_API_KEY`, wired through `APIKeyManaging`), for subscribers who don't run the CLI — e.g. Kimi Code driven through Claude Code;
- real plan names: Kimi's public `ListGoods` catalog maps the enum onto the tier users actually see (`LEVEL_INTERMEDIATE` → "Allegretto"; note `LEVEL_STANDARD` is the *top* tier, not a middle one). Title-casing the enum shows "Intermediate" instead;
- a tolerant per-field decode of the CLI's credential file: with a strict decode, one unexpectedly-typed field reports a signed-in user as broken credentials, and a file holding a valid `refresh_token` with an empty `access_token` is rejected outright;
- typed-container checks in the mapper (a `limits` that isn't an array is a malformed response, not "No usage data").

One small thing: in `windowDurationMs`, `unit.contains("SECOND")` also matches `TIME_UNIT_MILLISECOND`. Harmless today because the session window is matched by exact duration — but it's a 1000x error waiting for the day it isn't.

Happy to help get this over the line either way. On the maintenance question that sank both earlier attempts: @bzqzheng and I have both offered ongoing test access, so if you're a subscriber too, that's three of us.

robinebers:
Hi, thanks but we do not accept new model providers unless there is lots of proven demand and we have access to a plan to test this with.

shrimpwtf:
> Hi, thanks but we do not accept new model providers unless there is lots of proven demand and we have access to a plan to test this with.

You've been offered a account to use multiple times by the community, it's incredibly simple to get usage out of, and it's one of the most popular models to be using right now. This is completely stupid. So much for being 'open'

Hugofsp93:
Hey! There is a lot of benchmarks showing that Kimi K3 it is now close to claude, codex, etc. Please, try to reconsider @robinebers !

arobison203:
The best part here is that #791 is approved. So we are just closing PRs that implement an approved issue. 

God, I love OSS sometimes.

robinebers:
> The best part here is that #791 is approved. So we are just closing PRs that implement an approved issue.
> 
> God, I love OSS sometimes.

You're right, and I removed the label. Thanks

arobison203:
> > The best part here is that #791 is approved. So we are just closing PRs that implement an approved issue.
> > God, I love OSS sometimes.
> 
> You're right, and I removed the label. Thanks

Terrible. I'll just use software that actually wants people to implement features.

DevOpsCraftsman:
This is completely rubbish, incredible...
Someone here as a good alternative to openusage with support for kimi moonshotAI?
(or open to sane contributors at least)

NalinGHub:
wow these maintainers are insufferable

PR #1060, opened 2026-07-29
Fix background CPU: gate periodic TimelineViews on popover visibility

shirok1:
## TL;DR
The app was burning steady background CPU (~5%) with just the tray icon showing and no window open, because several `TimelineView` clocks kept ticking after the popover was hidden. This gates them so they stop when the popover isn't visible.

## What was happening
- `StatusItemController` hides the panel via `panel.orderOut(nil)` without tearing down the hosted SwiftUI tree, so the view graph stays fully mounted and alive in the background.
- `PopoverFooter`'s 1s countdown, `WidgetRowView`'s 30s reset-tick, and `ICloudSyncSettingsSection`'s 60s "Updated Xm ago" label each used a raw `TimelineView(.periodic(...))` with no visibility gate.
- `NSHostingView` drives `.periodic` schedules with a plain `NSTimer`, not a display link tied to on-screen state — so hiding the panel with `orderOut` never paused these timers.
- Every tick forced a full `NSHostingView.layout()` pass (visible in an Instruments/`sample` trace as deep, repeated `sizeThatFits`/`StackLayout` recursion and `CA::Transaction::commit()` work), even with nothing on screen.

## What this changes
- Adds `VisibilityGatedPeriodicTimeline` in `PartyMode.swift` — the `.periodic` counterpart to the existing `VisibilityGatedTimeline` already used to gate the party-mode egg animation on `\.popoverIsVisible`.
- Switches all three ungated call sites (`PopoverFooter`, `WidgetRowView`, `ICloudSyncSettingsSection`) to it, so each clock structurally unmounts (and its timer actually stops) whenever the popover isn't on-screen, and remounts instantly on reopen.
- Corrects a stale comment in `WidgetRowView.swift` that incorrectly claimed `TimelineView` already paused itself while hidden.

## Heads-up
- Confirmed all three sites are descendants of the `.environment(\.popoverIsVisible, ...)` modifier set in `DashboardView`, so the gate applies correctly in every case (including the Settings screen, which shares the same view tree as the dashboard rather than a separate window).

## Tests
- `swift build` passes clean (only pre-existing, unrelated warnings).
- Manually diagnosed via an idle `sample` profile of the running app showing continuous layout/display-cycle work driven by `NSHostingView.setUpdateTimer` while no window was on screen; recommend re-sampling after this fix to confirm CPU drops to ~0% at idle.

Co-Authored-By: Claude Sonnet 5 <noreply@anthropic.com>

### DIFF
--- a/Sources/OpenUsage/Support/PartyMode.swift
+++ b/Sources/OpenUsage/Support/PartyMode.swift
@@ -15,12 +15,15 @@ extension EnvironmentValues {
     }
 }
 
-/// Whether the hosting popover is currently on-screen. The easter-egg animation loops read this to
-/// **mount** their `TimelineView(.animation)` clocks only while the popover is visible and drop to a
-/// static frame when it isn't, so a closed popover with the egg still active runs no display link and
-/// spends no CPU. Default `false`, so the windowless ShareCard export and any non-popover host never
-/// mount the loops. Seeded from `PopoverTransparencyStore.popoverShown`, which `StatusItemController`
-/// flips at its `showPanel`/`hidePanel` chokepoints.
+/// Whether the hosting popover is currently on-screen. The easter-egg animation loops, plus any clock-
+/// driven text (footer countdown, per-row reset ticks, iCloud "Updated Xm ago" labels), read this to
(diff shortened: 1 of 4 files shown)


shirok1:
Fixes #1061

PR #1062, opened 2026-07-29
Add Qwen Cloud Token Plan provider

Anrahya:
## Approved issue

_Draft PR — no approved issue yet. Will open one via the new-provider template and link it before requesting review._

## TL;DR

Adds a `qwen` provider that tracks Qwen Cloud Token Plan (individual edition) usage: the rolling 5-hour and weekly quota windows shown on the plan's billing page, with reset countdowns, plan tier, and an auto-renew warning.

## What was happening

OpenUsage had no Qwen provider — Token Plan usage was only visible on the qwencloud.com billing page. The plan's data is unreachable with API keys (they are inference-only; verified by probing the gateway), so the existing "read a credential the provider's CLI already stored" pattern could not apply: the billing API is browser-session-cookie-only, exactly the calls the billing page itself makes.

## What this changes

- New `Sources/OpenUsage/Providers/Qwen/` module (`ProviderRuntime`): auth store (`~/.config/openusage/qwen.json`, 0600), usage client (`home.qwencloud.com/tool/user/info.json` for the session probe + `sec_token`, then three `cs-data.qwencloud.com/data/api.json` calls: `usage` (required), `subscription` and `quota-config` (best-effort)), pure mapper verified against live-captured payloads.
- New `SessionManaging` protocol (sibling of `APIKeyManaging`) + a Sign In / Sign Out session card in the provider's Customize detail, and `QwenLoginController`: the app's first interactive sign-in — a WKWebView window on the persistent website data store, so after one login, later sign-ins silently recapture the live session. A hand-pasted cookie file works too (headless fallback).
- Widgets: Session (5h) and Weekly percent meters with reset countdowns; plan label with tier quota numbers; amber header warning when auto-renew is off.
- Registration in `ProviderCatalog`/`AppContainer`/`DefaultLayout`, arc-free `qwen.svg` mark, `docs/providers/qwen.md`, and tests.

## Heads-up

- **First interactive sign-in in the app.** The codebase ethos is "never ask the user for a token," but Qwen exposes no local credential to read (inference keys can't reach billing data — verified). The WebView captures only what the console itself sets, verifies the session with a live `info.json` call before persisting, stores cookies 0600, and never logs them. The provider is fully functional with just the config file, so the WebView layer is removable if unwanted.
- **Size:** ~1,552 insertions including tests (608 lines). If the 1,000-line policy applies, happy to split into an implementation+docs PR (~944) and a tests PR (~608) — guidance welcome.
- **Metric placement defaults** (enabled: session+weekly; pinned: session+weekly following the Z.ai precedent; both Always Visible) are provisional pending owner confirmation per AGENTS.md.
- **International (Singapore) edition only**; the region is carried in the config so other editions can follow. The console API is internal/unversioned, so decoding is strict (failures surface as categorized errors, never silent zero meters) and tests pin the real payload shapes.

## Tests

- `swift test`: 1,328 tests pass (3 pre-existing skips), including 40+ new tests — mapper tests against live-captured payloads (unknown-field tolerance, login-failure envelopes, clamping), provider tests with `RoutingHTTPClient`/`FakeFiles` (happy path, expired session, decode/HTTP/rate-limit/transport failures, best-effort degradation, `SessionManaging` round-trip), and pure-function tests for the cookie capture.
- Verified in a local `build_and_run.sh` build against a real Token Plan account: fresh install auto-enabled the provider via `hasLocalCredentials()`, refresh logged `qwen ok`, and the local HTTP API reported live Session/Weekly meters with correct reset times.

## Screenshots

Will attach before marking ready: menu-bar widgets, the Customize session card, and the sign-in window.

### DIFF
--- a/Sources/OpenUsage/App/AppContainer.swift
+++ b/Sources/OpenUsage/App/AppContainer.swift
@@ -18,6 +18,10 @@ final class AppContainer {
     /// `APIKeyManaging`. Each matching Customize provider detail shows an API Key section and writes
     /// changes through the capability. Empty when no installed provider needs a user key.
     let apiKeyProviders: [any APIKeyManaging]
+    /// Providers whose credential is a captured web session (currently Qwen Cloud), conforming to
+    /// `SessionManaging`. Each matching Customize provider detail shows a session card (Sign In /
+    /// Sign Out) beside the API Key card's slot. Empty when no installed provider uses sessions.
+    let sessionProviders: [any SessionManaging]
     /// Quota pace notification preferences (three independent triggers). Drives the Settings section
     /// and is read by `WidgetDataStore.evaluateNotifications`.
     let notificationSettings: NotificationSettingsStore
@@ -82,6 +86,7 @@ final class AppContainer {
         )
         let registry = WidgetRegistry.from(providers)
         let apiKeyProviders = providers.compactMap { $0 as? any APIKeyManaging }
+        let sessionProviders = providers.compactMap { $0 as? any SessionManaging }
         let enablement = ProviderEnablementStore()
         let notificationSettings = NotificationSettingsStore()
         let layout = LayoutStore(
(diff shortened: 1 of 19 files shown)


robinebers:
Hi, thanks but we do not accept new model providers unless there is lots of proven demand and we have access to a plan to test this with.

PR #1105, opened 2026-08-13
fix: restore Z.ai credit quota limits

davidarny:
## TL;DR

Restore Z.ai Session and Weekly meters after the quota API renamed percentage limits from `TOKENS_LIMIT` to `CREDIT_LIMIT`.

## What was happening

- Z.ai refreshes completed successfully, so the log reported `zai ok`.
- Current GLM Coding Lite responses label both percentage quota windows as `CREDIT_LIMIT`.
- The mapper only recognized the older `TOKENS_LIMIT` name and returned a successful no-data snapshot, leaving both meters blank.

## What this changes

- Recognizes current `CREDIT_LIMIT` and legacy `TOKENS_LIMIT` entries as the same percentage quota shape.
- Keeps the existing window-based Session and Weekly classification.
- Adds an anonymized live-response regression fixture for the current GLM Coding Lite payload.
- Updates the Z.ai provider documentation with both response names.

## Heads-up

- These are undocumented endpoints used by Z.ai's own subscription UI, so the live-response fixture intentionally preserves only the structural fields consumed by the mapper.

## Tests

- `swift test --filter ZAI` — 42 passed.
- `swift test` — 1,278 XCTest tests passed, 3 skipped; 3 Swift Testing tests passed.
- `CONFIG=debug ./script/build_and_run.sh verify` — dev app rebuilt, launched, and verified.
- Live API + UI — Session rendered as 100% left and Weekly as 2% left instead of `No data`.

## Screenshots

Verified in the dev build with the production settings copied into the dev defaults domain:

Z.ai Session and Weekly limits

### DIFF
--- a/Sources/OpenUsage/Providers/ZAI/ZAIUsageMapper.swift
+++ b/Sources/OpenUsage/Providers/ZAI/ZAIUsageMapper.swift
@@ -2,8 +2,10 @@ import Foundation
 
 /// Builds metric lines from the Z.ai `/api/monitor/usage/quota/limit` payload and the plan name from
 /// `/api/biz/subscription/list`. Ports and extends the legacy Tauri plugin's mapping:
-/// - a `TOKENS_LIMIT` entry whose window is sub-daily (`unit: 3`, hours) is the 5-hour session meter,
-/// - a `TOKENS_LIMIT` entry whose window is multi-day (`unit: 6`, weeks) is the weekly meter,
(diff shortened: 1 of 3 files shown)


robinebers (approved):
Looks correct. The live GLM Coding Lite fixture matches the renamed `CREDIT_LIMIT` shape, Session/Weekly still split by window, and the old `TOKENS_LIMIT` path stays intact. Docs and the regression test are enough.

Codex’s type/name note is real but theoretical — current payloads use `type`, and `name` still works when `type` is missing. Not blocking.

PR #1111, opened 2026-08-14
perf: reduce scrolling update overhead

davidarny:
## TL;DR

Reduces avoidable SwiftUI and SVG work while the dashboard scrolls. Keeps reference-backed reorder geometry and cached provider SVG paths, while restoring row-local reorder gestures instead of attaching one simultaneous drag gesture to the whole dashboard.

**This PR remains draft. The Instruments results are directional hitch measurements, not an exact FPS claim.**

## What was happening

- Reorder geometry preferences replaced a SwiftUI state dictionary as scrolling changed row positions, invalidating the surrounding lists.
- Provider SVG path data was parsed again while icon shapes rendered.
- The dashboard-wide simultaneous drag recognizer also participated in ordinary scroll input and gave some of the improvement back.

## What this changes

- Keeps live reorder frames in a reference-backed store so geometry updates do not invalidate the entire list.
- Parses each provider SVG path and its bounds once, then reuses the cached values.
- Restores provider and metric reorder gestures to their own header or row.
- Preserves the existing provider and metric reorder paths in both the dashboard and Customize views.

## Heads-up

- The benchmark-only self-opening hook and scroll driver are not committed to the product branch.
- Automated runs used a release build, a fresh process, the Animation Hitches template, a 15-second trace, and the same 1,440-event pixel-scroll sequence.
- The scripted baseline recorded 510.8 ms hitch time/s. The former dashboard-wide gesture version recorded 373.3 and 384.7 ms/s. The product source now committed in this PR is byte-identical to the three-run candidate, which recorded 321.4, 41.4, and 174.2 ms/s.
- Candidate variance remains high, so these results support the code-path choice but do not establish a stable absolute frame rate.

## Tests

- `swift test --filter ReorderGeometryTests` — 3 tests passed, 0 failures.
- `swift test` — 1,289 tests passed, 3 skipped, 0 failures; the Swift Testing run also passed 3 tests.
- `./script/build_and_run.sh build` — release bundle built successfully (using the repository's prebuilt icon fallback and ad-hoc signing in this environment).
- Deterministic Instruments matrix completed for baseline, state-only, combined, SVG cache, local coordinates, idle geometry, LazyVStack, scroll hit testing, and the former PR HEAD.

### DIFF
--- a/Sources/OpenUsage/Support/MenuBarIcon.swift
+++ b/Sources/OpenUsage/Support/MenuBarIcon.swift
@@ -17,7 +17,7 @@ enum MenuBarIcon {
         let renderer = ImageRenderer(
             // Smaller inset than the provider default so the brand gauge keeps its prior menu-bar size
             // (its art already carries ~8% margin inside the source viewBox).
-            content: ProviderIconShape(pathData: mark.path, inset: 0.08)
+            content: ProviderIconShape(mark: mark, inset: 0.08)
                 .fill(Color.black)
                 .frame(width: side, height: side)
         )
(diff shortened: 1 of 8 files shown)


davidarny:
@robinebers @validatedev so you can check this build vs main branch. scrolling should be better in terms of FPS.

PR #1151, opened 2026-08-24
Claude accounts 7: explicit account runtime planning

robinebers:
## TL;DR

Replace special-case Claude account assembly with a single explicit runtime plan for every verified account.

## What was happening

- The default Claude card and additional account cards followed different assembly paths.
- Login swaps, account renames, source changes, and organization overlap made those paths difficult to keep consistent.

## What this changes

- Separate account observation, source planning, Cowork partitioning, persistent reconciliation, and runtime binding.
- Give every discovered Claude account exactly one verified runtime and stable card identity.
- Share the same runtime plan between the menu-bar app and the command-line usage reader.
- Replace overlapping historical assembly cases with focused account ownership and login-switch regressions.

## Heads-up

- Six changed files; the large deletion is the removal of the old special-case assembly and redundant tests.

## Tests

- 53 focused account assembly, discovery, scoped credential, login isolation, and migration tests passed.

---

> [!NOTE]
> **Medium Risk**
> This rewires how Claude credentials, Desktop fallback, and spend logs are bound to cards. A planning bug could mix accounts or drop usage, but the change is assembly-only and covered by focused ownership/login-switch tests.
> 
> **Overview**
> Replaces the special-case default-vs-extra Claude card path with one explicit runtime plan. Every verified account now gets a single bound card; the hardcoded default runtime exists only when discovery cannot prove any Claude account.
> 
> Assembly is split into observe → plan → Cowork partition → reconcile → bind. Credential ownership is decided at bind time (Desktop access, unscoped keychain fallback, expected identity), and spend roots attach only after identity is canonicalized. Ambiguous org-less identities, unidentified Cowork sandboxes, and truncated scans are quarantined instead of folding onto the default card.
> 
> `ProviderCatalog` and the CLI `UsageReader` consume the same `ClaudeRuntimePlan`, so login swaps keep the original `claude` card on its moved config dir or Desktop org while the new default-home login gets its own runtime.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 336cfab6a09f21c1ba46e89bb90618e928c76812. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/Sources/OpenUsage/App/AppContainer.swift
+++ b/Sources/OpenUsage/App/AppContainer.swift
@@ -77,14 +77,7 @@ final class AppContainer {
         let accountAssembly = ProviderAccountAssembly.make(accountsStore: accounts, waitsForLoginShell: true)
         self.accounts = accounts
 
-        let providers = ProviderCatalog.make(
-            claudeCards: accountAssembly.claudeCards,
-            defaultClaudeExtraLogRoots: accountAssembly.defaultClaudeExtraLogRoots,
-            defaultClaudeDisplayName: accountAssembly.defaultClaudeDisplayName,
-            defaultClaudeCardID: accountAssembly.defaultClaudeCardID,
-            defaultClaudeCoworkRoots: accountAssembly.defaultClaudeCoworkRoots,
-            defaultClaudeOrganization: accountAssembly.defaultClaudeOrganization
-        )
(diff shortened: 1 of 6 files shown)


robinebers:
Consolidated into #1146 so the multi-account implementation remains reviewable as four cohesive PRs.

PR #1206, opened 2026-09-03
fix(antigravity): restore local spend from all conversation stores and price every logged model

### LINKED ISSUE #1202: [Antigravity] Local spend and usage history not populating due to protobuf timestamp schema drift and SQLite WAL -readonly failure
### Description

In OpenUsage 0.7.10 (following PR #1139), Antigravity quota meters update as expected, but Antigravity local usage logs, Usage Trend, Today, Yesterday, and Last 30 Days spend metrics remain empty ($0.00 / 0 tokens).

Investigating the local conversation databases revealed two separate issues:
1. **Protobuf schema drift in AGY 1.1.x+**: `AntigravityProtoDecoder` strictly expects an embedded wall-clock timestamp inside `gen_metadata.data` (`field 1 -> field 9 -> field 4 -> field 1`). In recent AGY versions (from mid-August 2026 onward), AGY dropped `field 4` from `field 9`. Because this field is treated as mandatory, 100% of generation records created by current AGY versions fail decoding silently.
2. **SQLite WAL-mode failure with `-readonly`**: `SQLiteCLIAccessor.queryValue` executes `/usr/bin/sqlite3` with the `-readonly` flag. Because Antigravity conversation databases run in SQLite WAL (Write-Ahead Logging) mode, opening with `-readonly` fails with `unable to open database file (14)` (`SQLITE_CANTOPEN`) when shared memory (`-shm`) cannot be accessed or initialized read-only.

### Steps to Reproduce

1. Install and launch OpenUsage (v0.7.10).
2. Enable the Antigravity provider and run prompts in Antigravity CLI (v1.1.x+).
3. Open the OpenUsage menu bar panel and inspect the Antigravity card.
4. Notice that Quotas (Session/Weekly) update properly, but Usage Trend and Today/Yesterday/30d spend show 0 tokens / $0.00.
5. Inspect `~/Library/Logs/OpenUsage/OpenUsage.log` and observe:
   `[WARN] [plugin:antigravity] usage query failed for [PATH]: Parse error in 7th command line argument: unable to open database file (14)`

### Expected Behavior

OpenUsage reads `~/.gemini/antigravity-cli/conversations/*.db`, decodes generation events from `gen_metadata`, and populates the Usage Trend chart and Today / Yesterday / Last 30 Days spend tiles.

### Actual Behavior

Antigravity spend metrics remain empty ($0.00 / 0 tokens). All generation events from current AGY versions are dropped.

### OpenUsage Version

0.7.10

### macOS Version

macOS 15+

### Mac Type

Apple Silicon

### Additional Context

- **Timestamp Source Alternative**: While `gen_metadata` no longer embeds the wall-clock timestamp in `field 9.4`, the database's `steps` table maintains a 1-to-1 index with `gen_metadata`. In `steps.metadata`, `field 1` is a standard protobuf timestamp (`field 1 = epoch seconds varint`). Joining `steps.metadata` in `AntigravityDbUsageScanner` provides the precise execution timestamp, with the database file's modification time serving as a secondary fallback if `steps` is absent.
- **WAL Concurrency Handling**: In `SystemClients.swift`, falling back to opening without `-readonly` when SQLite returns error 14 resolves the WAL shared-memory conflict. Because `databaseExists` is already validated before calling the CLI runner, this fallback does not risk creating unintended files on disk.
- **Verified Impact**: Testing this resolution across local conversation databases successfully restored all missing generation events and accurately populated daily spend history and trend charts that previously showed 0.

### LINKED ISSUE #1215: Antigravity: `gemini-pro-default` / `gemini-default` placeholder IDs and model-less records show as unknown models
## TL;DR

Antigravity's local spend tile shows the "Unknown models found" warning for two reasons: the CLI writes placeholder model IDs (`gemini-pro-default`, `gemini-default`) that no alias rule matches, and some `gen_metadata` rows carry no model field at all but still decode as generation events.

## What is happening

- When the default/auto model is selected, Antigravity stores `gemini-pro-default` or `gemini-default` in `gen_metadata.data` field 19 instead of a real slug. The actual model is written alongside it in field 21 as a display name, e.g. `gemini-pro-default` → "Gemini 3.1 Pro (High)", `gemini-default` → "Gemini 3.5 Flash (Low)" / "Gemini 3.5 Flash (Medium)" / "Gemini 3.6 Flash (High)".
- `AntigravityProtoDecoder` reads only field 19, and `pricing_supplement.json` has no alias rule for `-default` slugs, so these events stay unpriced.
- Some rows have a different shape: field 1 is the system prompt text and field 2 holds messages, with no field 19. They still contain a field 4 with a nonzero varint, which the decoder reads as a token count, so it emits an event named "Unknown Antigravity Model". These look like prompt-context records rather than generations.

Sampled from one machine's `~/.gemini/antigravity-cli/conversations` (177 databases, ~3,700 records):

| Field 19 | Count | Field 21 seen |
|---|---|---|
| `gemini-pro-default` | 264 | Gemini 3.1 Pro (High), or absent |
| `gemini-default` | 182 | Gemini 3.5 Flash (Low/Medium), Gemini 3.6 Flash (High) |
| absent | 39 | none |

The `-default` rows are roughly 12% of records, so the 30-day spend is undercounted while they are unpriced.

## Proposed fix

1. In `AntigravityProtoDecoder.generationEvent`, when field 19 ends in `-default` (or is absent) and field 21 is present, use the field 21 display name as the model.
2. Add alias rules in `pricing_supplement.json` for display names with effort suffixes, e.g. `(?i)^Gemini 3\.1 Pro \((?:Low|Medium|High|Auto[^)]*)\)$` → `gemini-3.1-pro-preview`, and the same for 3.5/3.6/3.7/3.8 Flash. Today only the `(Auto…)` form is covered.
3. Decide whether records with neither field 19 nor field 21 should be skipped rather than reported as "Unknown Antigravity Model". Skipping hides them; keeping them preserves the current "absent model stays visibly unpriced" contract.

## Heads-up

- `gemini-pro-default` rows without field 21 (31 of 264 here) still cannot be resolved; they should keep the warning.
- Related: #1202 / #1206 (timestamp decoding for the same table).

Nabsku:
## Approved issue

Fixes #1202
Fixes #1215

## TL;DR

Antigravity's local spend tile was empty or partial for most users. This PR makes the conversation-log scanner decode newer records, open WAL-mode databases, read every Antigravity surface's store, and price the placeholder and tiered model IDs Antigravity logs, grouped by model family.

## What was happening

- Newer `agy` versions dropped the embedded wall-clock timestamp from `gen_metadata.data`, so every generation record failed to decode and the tile showed nothing.
- WAL-mode databases whose `-shm` sidecar cannot be opened read-only fail under `sqlite3 -readonly` with "unable to open database file (14)", so those conversations were skipped with a warning.
- Only `~/.gemini/antigravity-cli/conversations` was scanned. The Antigravity IDE, the Antigravity 2.0 app, and ACP sessions each keep their own store, so their usage was never counted.
- With the model picker on its default, Antigravity logs `gemini-default` / `gemini-pro-default` as the model ID and the served model only as a display label ("Gemini 3.1 Pro (High)"). No alias rule matched either, so about 12% of records were unpriced and the tile showed "Unknown models found".
- Prompt-context bookkeeping rows with no model and only a system-prompt token count decoded as generations and surfaced as "Unknown Antigravity Model".
- The breakdown listed every logged ID as its own row (`gemini-3.1-pro-low`, `gemini-3.7-flash-tiered`, `gemini-3.7-flash-exp-a`) even though they bill at the base model's rate.

## What this changes

- **Timestamps:** the decoder takes the correlated `steps.metadata` blob and uses its Timestamp when the embedded timing is absent. Records with neither are dropped rather than attributed to the database's modification day. The scanner probes for the `steps.metadata` column per database instead of matching error text, and re-probes a store that lacked it so one that gains step timestamps is re-read without a restart.
- **WAL reads:** `SQLiteCLIAccessor` retries any failed `-readonly` open with `PRAGMA query_only = ON`, so the fallback connection still cannot write.
- **Stores:** the scanner discovers every `~/.gemini/antigravity*/conversations` directory, so a new Antigravity surface needs no release. Symlinked or repeated stores are scanned once. An unreadable root, or a `~/.gemini` that cannot be listed, is reported through the read-failure path instead of aborting the scan or reading as no usage.
- **Model resolution:** the decoder exposes both the internal ID and the display label. The scanner prices placeholder rows by the label first and falls back to the ID, folds `-tiered` into the base model, and only reports a row as unknown when nothing prices. Bookkeeping rows produce no event.
- **Breakdown rows** group by canonical pricing family, as Cursor's do, with the catalog's `-preview` suffix dropped: effort variants, display labels, placeholder IDs, and experiment slots all count under one `gemini-3.1-pro` or `gemini-3.7-flash` row. Names no alias rule knows keep their raw text.
- **Alias rules:** display labels with any parenthetical qualifier for Gemini 3.1 Pro, 3.5 to 3.8 Flash, and Claude Opus/Sonnet 4.6; `gemini-pro-default` and `gemini-pro-agent` as Gemini 3.1 Pro; Flash `-exp-[a-z]` experiment slots; the `claude-opus-4-6` / `claude-sonnet-4-6` slug spellings. `gemini-default` is deliberately unaliased because its label has pointed at different Flash generations over time.
- **Docs:** `docs/providers/antigravity.md` and `docs/pricing.md` describe the stores, placeholder handling, tiered IDs, and row grouping.

## Heads-up

- Users who ran the IDE or the 2.0 app will see Antigravity spend jump on the first refresh after this ships, because that history was never counted.
- Grouping by family hides the Low/High effort split. Cursor keeps that as per-row variants; the accumulator Antigravity shares with Claude and Codex has no variant support, so that is a follow-up if anyone misses it.
- The query-only retry launches a second `sqlite3` process per batch on affected databases. Caching the working mode per path is a follow-up.

## Tests

- Verified against 5,166 records across all four local stores on one machine. The IDE and 2.0 databases reproduce the WAL open failure and read correctly through the retry. After the change the unknown-models warning is empty, 30-day spend prices in full, and the breakdown shows six family rows.
- Antigravity tests are split by concern (`AntigravityProtoDecoderTests`, `AntigravityDbSchemaTests`, `AntigravityDbUsageScannerTests`, shared fixtures in `AntigravityTestSupport`). They cover step-metadata timestamps, the no-timestamp drop, a real WAL-mode database, a `steps` table without `metadata`, a store that gains step timestamps, store discovery and symlink de-duplication, an unlistable `~/.gemini`, placeholder and tiered resolution, the priced-ID fallback, family grouping, and bookkeeping rows. `ModelPricingTests` covers the new alias rules; `CredentialCacheIntegrityTests` pins the retry arguments.

## Screenshots

Not applicable (data ingestion for existing spend tiles and charts).


validatedev:
@codex review

Sources/OpenUsage/Providers/Antigravity/AntigravityDbUsageScanner.swift
```diff
@@ -150,8 +167,20 @@ actor AntigravityDbUsageScanner {
...
-            guard let payload = try sqlite.queryValue(path: path, sql: Self.dataSQL(after: cached.lastIndex)) else { break }
+            let sql = usesLegacySQL ? Self.legacyDataSQL(after: cached.lastIndex) : Self.dataSQL(after: cached.lastIndex)
+            let payload: String?
+            do {
+                payload = try sqlite.queryValue(path: path, sql: sql)
+            } catch {
+                if !usesLegacySQL {
+                    usesLegacySQL = true
+                    continue
+                }
```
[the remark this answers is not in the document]
Nabsku: Addressed in b10efd0. Restricted the fallback to missing-schema SQLiteErrors and added test coverage. Thanks!

Sources/OpenUsage/Providers/Antigravity/AntigravityDbUsageScanner.swift
```diff
@@ -165,8 +194,12 @@ actor AntigravityDbUsageScanner {
                     cached.sawOversizedBlob = true
                     continue
                 }
+
+                let stepTimestamp = row.stepHex.flatMap(Self.bytes(fromHex:)).flatMap(AntigravityProtoDecoder.timestamp(fromStepMetadata:))
+                let fallbackTimestamp = stepTimestamp ?? Int64(cached.fingerprint.latestModification.timeIntervalSince1970)
```
robinebers: **[P2] Avoid assigning a mutable database timestamp to historical events**

When an event has neither embedded timing nor a usable step timestamp, this fallback treats the database's latest modification time as its generation time. That value changes when the conversation database is updated, while the warm scanner retains previously cached event dates. I reproduced this with a real SQLite database containing one timestamp-less generation and no `steps` table: after scanning it, advancing the database modification time by one day, and creating a fresh scanner, the same 15 tokens moved from September 9 to September 10; the existing scanner still reported September 9. This makes Today/Yesterday and daily history depend on whether OpenUsage has restarted, and can bring older usage into the current reporting window. Use a stable, authoritative per-event timestamp, or explicitly handle unavailable timing without attributing those records to an invented day. Add regression coverage for modification followed by restart.

Sources/OpenUsage/Providers/Antigravity/AntigravityDbUsageScanner.swift
```diff
@@ -108,7 +108,24 @@ actor AntigravityDbUsageScanner {
 
     /// `CASE` prevents SQLite from expanding oversized blobs, while the inner `LIMIT` bounds the
     /// maximum hex payload returned by any subprocess to roughly `batchSize * maximumBlobBytes * 2`.
+    /// Modern Antigravity conversation DBs omit timestamps from `gen_metadata.data` and record them
+    /// in `steps.metadata`, which is correlated by `idx`.
     static func dataSQL(after index: Int) -> String {
+        """
+        SELECT json_group_array(json_object('index', g.idx, 'hex',
+            CASE WHEN length(g.data) <= \(maximumBlobBytes) THEN hex(g.data) ELSE NULL END,
```
robinebers: **[P2] Update the batch-bounds regression test with the SQL aliases**

The query now uses `length(g.data)` and `hex(g.data)`, but `testBatchSQLBoundsRowsAndSkipsOversizedBlobsBeforeHexExpansion` in `AntigravityDbUsageScannerTests.swift` still asserts that the query contains `length(data)` and `hex(data)`. Running the selected tests on this commit produces deterministic failures at lines 402 and 403. Update these assertions alongside the query while preserving coverage of the row and blob-size bounds; a real SQLite fixture would also verify that both generation data and step metadata obey the intended limits.

robinebers (changes requested):
## AI Review — Codex

AI-generated review, posted on behalf of @robinebers.

Requesting changes on `b10efd0` for two findings detailed inline: the file-modification-time fallback makes daily usage change after a restart, and the SQL change breaks an existing regression test.

Validation: the code builds, and 75 of 76 selected Antigravity and credential-system-client tests pass. The failing test has two assertion failures. Additional temporary tests against real SQLite databases confirm that the modern step-timestamp path works and reproduce the restart-dependent date shift when timestamps are absent. The temporary tests were removed after review.

The PR remains unmerged.

Sources/OpenUsage/Providers/Antigravity/AntigravityDbUsageScanner.swift
validatedev: Addressed in 1193d29. The modification-time fallback is removed: the `steps.metadata` timestamp is now the only fallback, and events with neither an embedded nor a step timestamp are skipped instead of being attributed to a day. Added `testEventsWithoutAnyTimestampAreNotAttributedToDatabaseModificationDay`, which scans a real SQLite fixture, advances the file's mtime by a day, rescans with a fresh scanner, and asserts identical daily results (also exercises the legacy path without a `steps` table). The test fails on the previous code and passes now.

Sources/OpenUsage/Providers/Antigravity/AntigravityDbUsageScanner.swift
validatedev: Addressed in 1193d29. The assertions now check `length(g.data)` / `hex(g.data)`, plus a new assertion that the `step_hex` subquery bounds `steps.metadata` with the same `maximumBlobBytes` limit. `legacyDataSQL` is covered too (cursor, `LIMIT`, blob bounds, no `steps` reference). `swift test --filter Antigravity`: 74 tests, 0 failures.

validatedev:
@codex review

Sources/OpenUsage/Providers/Antigravity/AntigravityDbUsageScanner.swift
```diff
@@ -36,37 +36,48 @@ actor AntigravityDbUsageScanner {
...
+    /// `agy` CLI, the IDE, the Antigravity 2.0 app, and ACP-hosted sessions. Discovering them by name
+    /// means a new surface is picked up without a release.
+    /// A missing `~/.gemini` means Antigravity was never installed and yields no stores; any other
+    /// listing failure (a permission problem, an I/O error) is surfaced instead of read as "no usage".
+    static func conversationsDirectories(underGeminiHome geminiHome: String) throws -> [String] {
+        try listDirectoryNames(at: geminiHome)
+            .filter { $0.hasPrefix("antigravity") }
+            .sorted()
+            .map { geminiHome.trimmingTrailingSlashes + "/" + $0 + "/conversations" }
+            .filter { FileManager.default.fileExists(atPath: $0) }
```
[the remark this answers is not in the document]
validatedev: Fixed in 9aeb4fde: the existence filter is gone, so every `antigravity*/conversations` candidate goes through the listing, which treats a missing directory as empty and records any other failure for the read-failure reporter. Added a regression test with an unreadable `conversations` entry.

validatedev (approved):

PR #1186, opened 2026-08-29
fix(opencode): support OpenCode 2 session_message schema and credential table

arpankanwer:
## Approved issue

Fixes #1124

## TL;DR

OpenUsage failed on OpenCode 2 beta — "Couldn't read OpenCode's local database" and no Go meters — because the scanner still hit the v1 `message` table / `auth.json` only. This unions both `message` + `session_message` tables and falls back from `auth.json` to the SQLite `credential` table.

## What was happening

OpenCode 2 (1.18.x beta, current stable) changed local storage:

* `message` (fields `role`, `providerID`, `modelID`, `tokens.total`) → `session_message` (columns `type`, `seq`, `data` with `type='assistant'`, JSON `model.providerID`, `model.id`, `tokens.input/output/reasoning/cache.*` — no `total`)
* `~/.local/share/opencode/auth.json` → SQLite `credential` table (`integration_id='opencode-go'`, `value` JSON `{"key":"sk-..."}`)

OpenUsage only queried `FROM message` with `$.role` / `$.providerID` / `$.tokens.total` and only read `auth.json`, so on a v2 DB every scan threw `no such table: message` or returned empty, and `goAPIKey()` returned nil.

## What this changes

* **`OpenCodeUsageScanner`**: `dataSQL`/`probeSQL` now `UNION ALL` both tables with coalesced JSON paths
* **`OpenCodeAuthStore`**: try `auth.json` first, then fallback to `credential` table

## Tests

swift build → Build complete
swift test --filter OpenCode → 40 passed

Issue #1088, opened 2026-08-11
OpenCode card drops BYO-key provider usage (~26% of local spend)


robinebers:
Split out from a comment on #1076, where a user reported that their Kimi K3 usage isn't tracked in OpenCode either.

### Problem

`OpenCodeUsageScanner` only counts two providerIDs:

```swift
/// The OpenCode-hosted providerIDs we track: the Go subscription and the Zen pay-as-you-go gateway.
/// Both write an authoritative `cost`; other (BYO-key) providerIDs log `cost: 0` and are out of scope.
static let hostedProviderIDs = ["opencode-go", "opencode"]
```

**The stated justification is wrong.** OpenCode records real per-message costs for plenty of BYO-key providers — it prices them itself from models.dev. Measured against a real local `opencode.db`:

| | messages | cost | tokens |
|---|---|---|---|
| Counted (`opencode-go` + `opencode`) | 1672 | $32.28 | 138M |
| Dropped (everything else) | 1542 | $11.55 | 112M |

The dropped spend is real recorded cost, not zeros: OpenRouter $6.94, Vercel $2.71, Fireworks $1.86, MiniMax $0.03. So roughly a quarter of this machine's OpenCode usage is invisible on the card.

There are two distinct gaps behind the #1076 report:

1. **Scope** — non-hosted providerIDs are dropped even when OpenCode recorded a real dollar cost.
2. **Pricing** — OpenCode prices from models.dev, which has no Kimi K3 entry, so it writes `cost: 0`. The `openrouter/moonshotai/kimi-k3` rows on this machine are 10.4M tokens at $0. Including them without re-pricing would show tokens against $0.

The Go plan itself is clean and needs no change: its 31 zero-cost rows carry zero tokens, so Go always records real spend when there is real usage. This is purely a BYO-key problem.

### Why it isn't a one-line fix

Dropping the filter double-counts. OpenUsage already has an OpenRouter card reporting account-level billed dollars, which includes OpenCode's OpenRouter traffic, and the Cursor card covers `cursor-acp` rows. Inflating Total Spend would be a worse bug than the current undercount. Vercel, Fireworks, and MiniMax have no card of their own, so those are pure gain.

It also changes what the card *means*. `docs/providers/opencode.md` currently promises "OpenCode-hosted usage — the Go subscription and the Zen pay-as-you-go gateway".

### Proposed solution

[...]

PR #1243, opened 2026-09-10
fix(opencode): support OpenCode 2 session_message schema and credential table

### LINKED ISSUE #1124: OpenCode 2 (session_message schema) not detected - local database unreadable + missing auth.json
### Description
OpenUsage doesn't detect OpenCode after upgrading to OpenCode 2 (current stable `1.18.18` / `0.0.0-beta-17639` Go binary at `~/.opencode/bin/opencode`). Menu bar shows `Couldn't read OpenCode's local database` and `/v1/limits` returns `errors: [{providerId: "opencode", message: "Couldn't read OpenCode's local database..."}]`. After the DB workaround the Go meters reappear but `auth.json`-based detection also fails.

### Steps to Reproduce
1. Install OpenCode 2 (new Go/TUI) via `npm i -g @opencode-ai/cli` or `curl -fsSL https://opencode.ai/install | bash` -> binary `~/.opencode/bin/opencode` version `1.18.18`, data dir `~/.local/share/opencode/opencode.db`
2. Run a few sessions (`opencode run "hi"`), confirm `~/.local/share/opencode/opencode.db` has table `session_message` (not `message`) with rows where `json_extract(data,'$.model.providerID')='opencode-go'`
3. Open OpenUsage `0.7.10-beta.1` -> OpenCode card shows error. `curl http://127.0.0.1:6736/v1/limits` and `openusage --force` both show the error. `curl http://127.0.0.1:6736/v1/usage` shows no OpenCode spend tiles.
4. Check DB: `sqlite3 ~/.local/share/opencode/opencode.db "SELECT name FROM sqlite_master WHERE type='table';"` -> shows `session_message` etc, no `message`. Check auth: `ls ~/.local/share/opencode/auth.json` -> missing (key is now in `credential` table).

### Expected Behavior
OpenUsage detects OpenCode 2 local usage and shows:
- Go meters (Session/Weekly/Monthly from `https://opencode.ai/zen/go/v1/usage`)
- Spend tiles Today / Last 30 Days + Usage Trend from local SQLite (combined Go+Zen), same as before the upgrade.

### Actual Behavior
- `hasLocalCredentials()` returns false via `hasHostedUsage()` probe because `SELECT 1 FROM message ...` fails with `no such table: message`
- `scan()` throws `databaseUnreadable` for same reason
- `goAPIKey()` returns nil because `~/.local/share/opencode/auth.json` doesn't exist anymore; key is now in SQLite `credential` (`SELECT json_extract(value,'$.key') FROM credential WHERE ... LIKE 'sk-%'`)

### OpenUsage Version
0.7.10-beta.1 (523)

### macOS Version
15.5+ (Sequoia, Apple Silicon)

### Root Cause
`Sources/OpenUsage/Providers/OpenCode/OpenCodeUsageScanner.swift:42` and `Sources/OpenUsage/Providers/OpenCode/OpenCodePaths.swift:28` still target the OpenCode 1 schema:

* Table `message` -> renamed to `session_message` in OpenCode 2 (see `~/.local/share/opencode/opencode.db` schema: `session_message(id, session_id, type, seq, time_created, time_updated, data)`)
* JSON shape changed:
  - `$.role` -> column `type` (`assistant`/`user`)
  - `$.providerID` -> `$.model.providerID`
  - `$.modelID` -> `$.model.id`
  - `$.tokens.total` -> `COALESCE($.tokens.input,0)+COALESCE($.tokens.output,0)+COALESCE($.tokens.reasoning,0)+COALESCE($.tokens.cache.read,0)+COALESCE($.tokens.cache.write,0)`
  - `$.cost` stays at top level but is now `NULL` for user/error rows (filtered by `json_type IN ('integer','real')`)
* Auth moved from `~/.local/share/opencode/auth.json` (`{"opencode-go":{"key":"sk-..."}}`) to SQLite `credential` table. `OpenCodeAuthStore.swift:40` only reads the file.

Strings extracted from binary confirm:
```
SELECT 1 FROM message WHERE json_valid(data) AND json_extract(data,'$.role')='assistant' AND json_extract(data,'$.providerID') IN ('opencode-go','opencode') ...
SELECT json_group_array(json_array(time_created, json_extract(data,'$.cost'), COALESCE(json_extract(data,'$.tokens.total'),0), ...)) FROM message ...
```

### Proposed Fix
1. **Scanner**: probe for both tables, or create a compatibility VIEW. Suggested `dataSQL`/`probeSQL` that union both schemas or try `session_message` first:
```sql
-- fallback that handles both v1 and v2
SELECT ... FROM (SELECT * FROM message UNION ALL SELECT * FROM session_message) -- if both exist
```
Or, detection: if `session_message` exists, query it with transformed columns:
```sql
CREATE VIEW IF NOT EXISTS message AS
SELECT time_created,
  json_patch(json_patch(data, json_object('role', type)),
    json_object('providerID', json_extract(data,'$.model.providerID'),
                'modelID', json_extract(data,'$.model.id'),
                'cost', json_extract(data,'$.cost'),
                'tokens', json_object('total',
                  COALESCE(json_extract(data,'$.tokens.input'),0)+
                  COALESCE(json_extract(data,'$.tokens.output'),0)+
                  COALESCE(json_extract(data,'$.tokens.reasoning'),0)+
                  COALESCE(json_extract(data,'$.tokens.cache.read'),0)+
                  COALESCE(json_extract(data,'$.tokens.cache.write'),0))))
FROM session_message;
```
Better to make `OpenCodeUsageScanner` natively query `session_message` with the transformations, and keep `message` as fallback for old DBs. Also handle both `$.tokens.total` (v1) vs sum (v2) via `COALESCE(json_extract(data,'$.tokens.total'), sum(...))`.

2. **Auth**: make `OpenCodeAuthStore.goAPIKey()` try file first, then fallback to SQLite:
```swift
// if auth.json missing, open ~/.local/share/opencode/opencode*.db and query credential table
SELECT json_extract(value,'$.key') FROM credential WHERE json_extract(value,'$.key') LIKE 'sk-%' LIMIT 1
```
Respect `OPENCODE_DATA_DIR`/`XDG_DATA_HOME` via `OpenCodePaths.dataDirectory`.

3. **Tests**: extend `OpenCodeUsageScannerTests` to cover v2 row shape.

### Workaround (verified locally)
Created VIEW + synced auth.json:
```bash
KEY=$(/usr/bin/sqlite3 ~/.local/share/opencode/opencode.db "SELECT json_extract(value,'\$.key') FROM credential WHERE json_extract(value,'\$.key') LIKE 'sk-%' LIMIT 1;")
echo "{\"opencode-go\":{\"key\":\"$KEY\"}}" > ~/.local/share/opencode/auth.json; chmod 600 ~/.local/share/opencode/auth.json
/usr/bin/sqlite3 ~/.local/share/opencode/opencode.db "DROP VIEW IF EXISTS message; CREATE VIEW message AS SELECT time_created, json_patch(... ) AS data FROM session_message;"
```
After that `openusage --force` shows `errors: []`, `opencode` with Today `$0.20 · 21.4M` and Trend, and `/v1/usage` lines are correct. LaunchAgent with `WatchPaths` on `~/.local/share/opencode` keeps it persistent.

Happy to open a PR if you want - I have the SQL ready and tested against `/usr/bin/sqlite3` `3.51.0` (system) and `~/.local/share/opencode/opencode.db` with 271 hosted rows.

### Additional Context
- `/usr/bin/sqlite3` supports `json_patch` on macOS 15 (tested)
- Data dir resolution via `OPENCODE_DATA_DIR` should stay, just add DB fallback
- No other providers affected

arpankanwer:
## Approved issue

Fixes #1124

Supersedes #1186 (that PR could not be reopened — GitHub refuses while its head branch has been force-pushed after close, so this branch is a clean replacement, rebased onto current `main`).

## TL;DR

OpenUsage fails on OpenCode 2 — "Couldn't read OpenCode's local database" and no Go meters — because the scanners only query the v1 `message` table and the auth store only reads `auth.json`. This adds OpenCode 2 support: both OpenCode scanners union `message` + `session_message` with coalesced JSON paths, and credentials fall back from `auth.json` to the SQLite `credential` table.

## What was happening

OpenCode 2 changed local storage in two places:

* `message` (`role`, `providerID`, `modelID`, `tokens.total`) → `session_message` (columns `type`, `seq`, `data`; `type='assistant'` replaces `$.role`, with `$.model.providerID`, `$.model.id`, and `tokens.input/output/reasoning/cache.*` — no `total`)
* `~/.local/share/opencode/auth.json` → the SQLite `credential` table (`integration_id='opencode-go'`, `value` JSON `{"key":"sk-..."}`)

On a v2 database the scan hits `no such table: message` (or reads only stale pre-upgrade rows) and `goAPIKey()` returns nil, so the card reports an unreadable database and drops all Go meters.

Measured on a migrated local `~/.local/share/opencode/opencode.db`:

| | hosted rows |
|---|---|
| `message` (v1) | 24,893 |
| `session_message` (v2) | 21,823 |
| union | **46,716** |

Both tables persist after migration, so a v1-only read roughly halves the spend tiles and reports `0` for any day after the upgrade.

## What this changes

* **`OpenCodeUsageScanner`** — `dataSQL`/`probeSQL` union both tables with coalesced JSON paths: `COALESCE($.model.providerID,$.providerID)`, `COALESCE($.model.id,$.modelID)`, and `COALESCE($.tokens.total, input+output+reasoning+cache.read+cache.write)`. `message` keeps its `$.role='assistant'` filter; `session_message` accepts `assistant` plus completed compactions. 2.0.3 migrates legacy rows under their original IDs, so copies are deduplicated by ID across tables and channel databases instead of counted twice.
* **`OpenCodeCodexUsageScanner`** — same v2 blind spot from #1195 (it matched `$.providerID = 'openai'` on the v1 table only), so Codex OAuth usage recorded after an upgrade counted as nothing. Now unions both tables with the columns projected once outside the union. Tokens fall back to the per-bucket sum because OpenCode 2 dropped `$.tokens.total`. Completed compactions count with their own `status`-based condition. v2 rows older than the current OAuth credential are excluded: experimental builds zeroed every cost including paid traffic, so those rows may be paid history rather than subscription usage (v1 rows priced paid traffic correctly and are unaffected).
* **`OpenCodeAuthStore`** — reads the live `credential` table first (OpenCode 2 imports `auth.json` without deleting it, so a retained file key goes stale; the file stays the OpenCode 1 fallback), still throwing `credentialsUnreadable` on a malformed file. Every lookup takes the current row first (`active DESC, time_updated DESC, id DESC`, admitting NULL-flagged imports) instead of an arbitrary `LIMIT 1` row. Go keys stay scoped to `opencode-go`/`opencode`.
* **`OpenCodePaths`** — owns the `OpenCodeMessageTables` set and the `sqlite_master` probe every scanner runs first.
* **`docs/providers/opencode.md`, `docs/providers/codex.md`** — document both credential sources, both message tables, and the credential-table location for the Codex OAuth lookup.

* **`OpenCodeProvider`** — a failed Go meters request no longer hides local tiles: it logs and falls through to the scan, surfacing as an error only when there is nothing else to show.

## Heads-up

* No migration, no new dependency, and no change for existing installs: `auth.json` is still tried first and v1 rows still count.
* BYO-key scope is unchanged — still only `opencode-go`/`opencode` on the OpenCode card, so Total Spend cannot double-count against the OpenRouter or Cursor cards (#1088 is a separate, larger question).
* Each scanner probes `sqlite_master` once per database and builds its query from the tables that are actually present (v1 only, v2 only, or the union), so a database holding only one of them is read rather than reported unreadable. A file with neither table is skipped. This costs one extra lightweight query per database per refresh.
* I could not exercise the OpenCode 2 path in CI — there is no v2 fixture in the suite. Verification below is a real migrated database, plus the stub SQLite.

## Tests

```
swift build                   → Build complete
swift test                    → 1356 tests, 3 skipped, 0 failures
swift test --filter OpenCode  → 83 passed
```

New coverage: migrated-copy dedupe (same ID twice and across channel DBs counts once), completed-compaction counting on both scanners, current-row credential ordering, DB-beats-stale-file in both directions, v2 OAuth-age bound (pre-credential excluded, post-credential and v1 included), Go-failure-keeps-tiles at provider level, schema-less exclusion from failure counting on both scanners (failed usable plus skipped sibling still throws/returns nil; all-skipped stays an empty result), the credential-table fallback for `goAPIKey()` and `hasCodexOAuth()` (including auth.json winning when both exist, and an API-key credential not counting as OAuth), both fallback queries naming their integration (no unscoped `LIKE 'sk-%'`), the best-effort nil contract on credential DB failures, a v2 assertion on the Codex scanner SQL, and hermetic `databasePaths: { [] }` on the auth-store tests so they can no longer read this machine's real databases. The existing `testQuerySelectsOnlyCompletedOpenAIRows` asserted the substring `providerID') = 'openai'`, which only the v1 form could satisfy; it now asserts the coalesced predicate actually executed.

Manual checks against a real OpenCode 2 database, running the scanners' own SQL:

```
OpenCode card, 30 days, union over both tables → 18,098 rows · 1,188,825,621 tokens · $13.62
  message only (v1)                            →  9,941 rows ·   646,561,838 tokens ·  $7.36
  session_message only (v2)                    →  8,157 rows ·   542,263,783 tokens

Codex attribution, same predicate shape       → 384 in message, 213 in session_message, 597 unioned
credential lookup                             → returns the opencode-go sk- key
```

Before/after through the local HTTP API on the same machine:

```
before   Today  $0    ·  0 tokens      Last 30 Days   $7.36 · 646.6M
after    Today  $1.14 ·  97M tokens    Last 30 Days  $13.62 ·   1.2B
```

## Screenshots

Not applicable — no visual change.


Sources/OpenUsage/Providers/OpenCode/OpenCodeUsageScanner.swift
```diff
@@ -173,24 +173,40 @@ struct OpenCodeUsageScanner: Sendable {
...
-        LIMIT 1;
+        SELECT 1 FROM (
+          SELECT 1 as x FROM message
+          WHERE json_valid(data)
+            AND json_extract(data,'$.role') = 'assistant'
+            AND COALESCE(json_extract(data,'$.model.providerID'), json_extract(data,'$.providerID')) IN \(providerFilter)
+            AND json_type(data,'$.cost') IN ('integer','real')
+          UNION ALL
+          SELECT 1 FROM session_message
+          WHERE type = 'assistant'
```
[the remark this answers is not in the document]
arpankanwer: Fixed in 08cf190.

Both scanners now probe `sqlite_master` once per database and build the query from the tables that are actually present — v1 only, v2 only, or the union — and a database with neither is skipped rather than counted as a failure. `OpenCodeMessageTables` in `OpenCodePaths.swift` holds the set and the probe; the projection is written once per scanner and wrapped around whichever bodies are selected, so the three variants can't drift.

Worth recording what I could and couldn't confirm about the trigger, because it affects how likely this is to bite:

* A fresh database does **not** come up v2-only. `packages/core/src/database/migration.ts` builds an empty database with `schema.up(tx)` from `schema.gen.ts`, which creates **both** `message` (line 128) and `session_message` (line 170).
* An upgraded database keeps both: `message` is created by the earliest migration and **no migration drops it** (the only `DROP TABLE` in the v2 migration is `session_entry`).
* So on current OpenCode every database has both tables, and I could not reproduce the failure against a real install.

The scenario is still worth defending against, since #1124's own repro reports a database with no `message` table on 1.18.18, and the cost is one extra lightweight query per database per refresh. Verified against real and synthetic databases: probe `1|1` on a current install, `0|1` on a `session_message`-only file, and the correct variant runs in each case.

Sources/OpenUsage/Providers/OpenCode/OpenCodeCodexUsageScanner.swift
```diff
@@ -170,30 +173,46 @@ struct OpenCodeCodexUsageScanner: Sendable {
...
+            AND json_valid(data)
+            AND json_extract(data,'$.role') = 'assistant'
+            AND COALESCE(json_extract(data,'$.model.providerID'), json_extract(data,'$.providerID')) = 'openai'
+            AND json_type(data,'$.cost') IN ('integer','real')
+            AND json_extract(data,'$.cost') = 0
+            AND (json_type(data,'$.time.completed') IN ('integer','real')
+                 OR json_type(data,'$.finish') = 'text')
+          UNION ALL
+          SELECT time_created, id, data FROM session_message
+          WHERE time_created >= \(creationCutoffMs)
```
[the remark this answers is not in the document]
arpankanwer: Fixed in 08cf190 — the Codex attribution scanner got the same treatment: it probes `sqlite_master` per database and queries only the tables present, skipping a database that has neither instead of dropping every other database's rows. See the reply on the sibling comment for what I could and couldn't confirm about the v2-only scenario.

validatedev:
@codex review

Sources/OpenUsage/Providers/OpenCode/OpenCodeAuthStore.swift
```diff
@@ -1,21 +1,52 @@
...
+    var databasePaths: @Sendable () throws -> [String]
+
+    /// Credential lookups against the OpenCode 2 `credential` table. The Go key tries its own
+    /// integration first, then any `sk-` key as a last resort (mirrors the `opencode-go` → `opencode`
+    /// fallback in the issue workaround). The Codex lookup reads the whole `value` object because the
+    /// OAuth fields live beside `type`, not under a single key.
+    private static let credentialSQLGoKey =
+        "SELECT json_extract(value,'$.key') FROM credential WHERE integration_id = 'opencode-go' AND json_extract(value,'$.key') LIKE 'sk-%' LIMIT 1;"
+    private static let credentialSQLAnyKey =
+        "SELECT json_extract(value,'$.key') FROM credential WHERE json_extract(value,'$.key') LIKE 'sk-%' LIMIT 1;"
```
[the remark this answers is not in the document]
arpankanwer: Fixed in f71527d — good catch, that was a real disclosure bug.

The fallback was `opencode-go`, then an unscoped `LIKE 'sk-%'`, so a database with no Go row but a BYO key (e.g. OpenAI) would return that key and `goAPIKey()` would send it as a Bearer token to the usage endpoint. It now tries `opencode-go`, then `opencode`, and nothing else. Added `testCredentialFallbackQueriesAreScopedToOpenCodeIntegrations` asserting both queries name their integration.

Sources/OpenUsage/Providers/OpenCode/OpenCodeAuthStore.swift
```diff
@@ -68,4 +120,28 @@ struct OpenCodeAuthStore: Sendable {
...
+            return nil
+        }
+        for path in paths {
+            do {
+                if let value = try sqlite.queryValue(path: path, sql: sql)?
+                    .trimmingCharacters(in: .whitespacesAndNewlines).nilIfEmpty {
+                    return value
+                }
+            } catch {
+                continue
```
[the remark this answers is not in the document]
arpankanwer: Fixed in f71527d.

`credentialValue` now distinguishes the two failure kinds: a missing `credential` table (pre-OpenCode-2 database) still skips silently, while anything else — locked, I/O, permissions, unreadable data directory — is logged via `AppLog.warn`. It still returns nil rather than throwing, so the Go path keeps reporting the scanner's own `databaseUnreadable` instead of a misleading credential error; the Codex path now at least leaves a log trail instead of vanishing silently.

One deliberate trade-off: this logs per refresh while a failure persists, where the scanners edge-trigger. A locked OpenCode database is rare enough that loudness wins over thrift here, per the fail-loud rule. Added `testCredentialDatabaseFailuresReturnNilInsteadOfThrowing` pinning the best-effort nil contract.

validatedev:
@codex review

Sources/OpenUsage/Providers/OpenCode/OpenCodeUsageScanner.swift
```diff
@@ -75,7 +75,10 @@ struct OpenCodeUsageScanner: Sendable {
         for path in paths {
             checked.insert(path)
             do {
-                if let json = try sqlite.queryValue(path: path, sql: Self.dataSQL(cutoffMs: cutoffMs)) {
+                // A database with no message tables has no usage to read; skip it instead of letting
+                // the query fail and paint the whole provider as unreadable.
+                guard let tables = try messageTables(in: path) else { continue }
```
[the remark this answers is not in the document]
arpankanwer: Fixed in 2335ce7 — you're right, and the arithmetic is exactly as you describe.

Skipped paths stayed in `checked`, so with one schema-less file plus one failed usable database, `failures.count (1) != checked.count (2)` and the scanner returned an empty scan instead of `databaseUnreadable`. Both scanners now count only databases that actually have message tables: skipped files vote on nothing, a probe throw still counts as a failure, and all-skipped stays an empty result rather than an error.

For the record on whose lines these are, since the review header names the merge commit: the loop scaffolding (`checked`/`failures` and the throw) is upstream's, but the skip that broke its counting is mine (08cf190) — so this one is on the PR, not on main.

Verified: new tests cover failed-usable-plus-skipped-sibling (throws/nil) and all-skipped (empty result) on both scanners. Full suite: 1344 passed, 3 skipped, 0 failures.

Sources/OpenUsage/Providers/OpenCode/OpenCodeUsageScanner.swift
```diff
@@ -168,29 +190,81 @@ struct OpenCodeUsageScanner: Sendable {
...
+
+    /// One table or two, the body is always a subquery so the projection above can read `FROM` it
+    /// uniformly. Callers never pass an empty set — a database with no message tables is skipped
+    /// before any SQL is built.
+    private static func rowsSource(_ tables: OpenCodeMessageTables, v1: String, v2: String) -> String {
+        let bodies = [
+            tables.contains(.v1) ? v1 : nil,
+            tables.contains(.v2) ? v2 : nil,
+        ].compactMap { $0 }
+        return "(\n" + bodies.map(indented).joined(separator: "\n          UNION ALL\n") + "\n        )"
```
robinebers, line 250: **[P1] Avoid Counting Migrated Messages Twice**

OpenCode 2.0.3 [copies legacy messages into session_message](https://github.com/anomalyco/opencode/blob/d44b52ca66b6bf69626c0384626d1a9cd9555977/packages/core/src/database/v1-migration.bun.ts#L613-L636), preserving their IDs, timestamps, costs and tokens while retaining the original message rows. This UNION ALL counts both copies, and the projection leaves out the IDs needed to deduplicate them. A real SQLite check returned $4 and 1,000 tokens for one migrated $2, 500-token request.

Carry message IDs through and deduplicate across both tables and channel databases, as the Codex scanner already does.

Sources/OpenUsage/Providers/OpenCode/OpenCodeAuthStore.swift
```diff
@@ -1,21 +1,51 @@
...
+
+    /// OpenCode 2 `credential`-table lookups. The Go-key fallback tries `opencode-go`, then `opencode`
+    /// — scoped to those two IDs so a BYO key (e.g. OpenAI) can never be mistaken for Go auth and sent
+    /// to the usage endpoint as a Bearer token.
+    static let credentialSQLGoKey =
+        "SELECT json_extract(value,'$.key') FROM credential WHERE integration_id = 'opencode-go' AND json_extract(value,'$.key') LIKE 'sk-%' LIMIT 1;"
+    static let credentialSQLOpencodeKey =
+        "SELECT json_extract(value,'$.key') FROM credential WHERE integration_id = 'opencode' AND json_extract(value,'$.key') LIKE 'sk-%' LIMIT 1;"
+    static let credentialSQLCodexOAuth =
+        "SELECT value FROM credential WHERE integration_id = 'openai' AND json_extract(value,'$.type') = 'oauth' LIMIT 1;"
```
robinebers: **[P2] Select the Current Database Credential Before Checking Its Type**

OpenCode 2 [keeps previous credentials as inactive when connecting a new account](https://github.com/anomalyco/opencode/blob/d44b52ca66b6bf69626c0384626d1a9cd9555977/packages/core/src/credential.ts#L100-L123). These unordered queries returned an old Go key despite a newer active key, and an inactive OAuth entry despite the current OpenAI credential being an API key. This can show the old account’s quotas, cause persistent authentication failures, or incorrectly pass the Codex OAuth guard.

Select the current credential using OpenCode’s active/time/ID ordering before checking its key or auth type. Filtering for OAuth before choosing the current row still resurrects an inactive account; requiring active=1 alone would miss imported credentials whose active value is NULL.

Sources/OpenUsage/Providers/OpenCode/OpenCodeAuthStore.swift
```diff
@@ -26,33 +56,54 @@ struct OpenCodeAuthStore: Sendable {
...
-        guard let object = try authObject() else { return nil }
-        guard let entry = object["opencode-go"] as? [String: Any],
-              let key = entry["key"] as? String
-        else { return nil }
-        return key.trimmingCharacters(in: .whitespacesAndNewlines).nilIfEmpty
+        if let object = try authObject(),
+           let entry = object["opencode-go"] as? [String: Any],
+           let key = entry["key"] as? String,
+           let trimmed = key.trimmingCharacters(in: .whitespacesAndNewlines).nilIfEmpty {
+            return trimmed
```
robinebers, line 76: **[P2] Do Not Let the Retained Legacy File Hide Updated Credentials**

OpenCode 2 [imports auth.json without deleting or updating it](https://github.com/anomalyco/opencode/blob/d44b52ca66b6bf69626c0384626d1a9cd9555977/packages/core/src/database/migration/20260805200742_import_legacy_credentials.ts#L46-L89). Subsequent login changes SQLite, but this early return keeps choosing the old file key. After rotating a key or switching accounts, users can receive persistent 401s or the old account’s quotas; logging in again only changes the database that this branch ignores. The real SQLite/file check returned sk-old despite the current database key being sk-new.

Resolve the applicable credential source, or try the current database credential after rejection of the retained legacy key.

Sources/OpenUsage/Providers/OpenCode/OpenCodeCodexUsageScanner.swift
```diff
@@ -170,30 +194,79 @@ struct OpenCodeCodexUsageScanner: Sendable {
...
+        SELECT time_created, id, data FROM session_message
+        WHERE time_created >= \(creationCutoffMs)
+          AND type = 'assistant'
+          AND json_valid(data)
+          AND COALESCE(json_extract(data,'$.model.providerID'), json_extract(data,'$.providerID')) = 'openai'
           AND json_type(data,'$.cost') IN ('integer','real')
           AND json_extract(data,'$.cost') = 0
           AND (json_type(data,'$.time.completed') IN ('integer','real')
-               OR json_type(data,'$.finish') = 'text');
+               OR json_type(data,'$.finish') = 'text')
```
robinebers: **[P2] Include the Model Requests Used for Compaction**

OpenCode 2 records automatic and manual conversation-summary requests as [type='compaction', status='completed', with their own model, cost and tokens](https://github.com/anomalyco/opencode/blob/d44b52ca66b6bf69626c0384626d1a9cd9555977/packages/schema/src/session-message.ts#L238-L265). Both scanners’ assistant-only filters leave out this consumption. These rows also lack the assistant completion markers required here, so widening the type filter alone is insufficient. A completed 151,000-token compaction fixture produced no usage under the actual query.

Include these request rows with the appropriate completion condition in both the Codex and OpenCode scanners.

Sources/OpenUsage/Providers/OpenCode/OpenCodeCodexUsageScanner.swift
```diff
@@ -170,30 +194,79 @@ struct OpenCodeCodexUsageScanner: Sendable {
...
+
+    private static func v2Rows(cutoffMs: Int, creationCutoffMs: Int) -> String {
+        """
+        SELECT time_created, id, data FROM session_message
+        WHERE time_created >= \(creationCutoffMs)
+          AND type = 'assistant'
+          AND json_valid(data)
+          AND COALESCE(json_extract(data,'$.model.providerID'), json_extract(data,'$.providerID')) = 'openai'
           AND json_type(data,'$.cost') IN ('integer','real')
           AND json_extract(data,'$.cost') = 0
```
robinebers, line 255: **[P2] Zero Cost Does Not Prove OAuth for Earlier V2 History**

On the published 1.18.25 experimental v2 path, the [runner records zero cost for every completed request](https://github.com/anomalyco/opencode/blob/cb7d8b2f5e44876ef98b661dc10590c915af3a9f/packages/core/src/session/runner/llm.ts#L333-L339), including [paid OpenAI API-key traffic](https://github.com/anomalyco/opencode/blob/cb7d8b2f5e44876ef98b661dc10590c915af3a9f/packages/core/src/session/runner/model.ts#L83-L87). After switching that connection to ChatGPT OAuth, those historical paid requests satisfy this new filter and enter Codex subscription totals. The actual scanner reproduced the incorrect attribution.

This applies while using that earlier v2 database or retaining it as a channel sibling. New 2.0.3 API requests record costs correctly; an ordinary same-file upgrade normally replaces the older experimental table. Treat the ambiguous earlier records separately instead of assuming cost=0 proves subscription usage.

Sources/OpenUsage/Providers/OpenCode/OpenCodeAuthStore.swift
```diff
@@ -26,33 +56,54 @@ struct OpenCodeAuthStore: Sendable {
...
+        }
+        for sql in [Self.credentialSQLGoKey, Self.credentialSQLOpencodeKey] {
+            if let key = credentialValue(from: sql) {
+                return key
```
robinebers: **[P2] Keep Zen Spend Tiles When the Go Request Fails**

Returning an ordinary Zen key here makes OpenCodeProvider.refresh() request Go meters before reading local history. A network or server failure returns an error immediately, hiding readable Zen spend tiles that previously worked offline. The provider check reproduced a network error and a missing Today tile solely because the Zen credential existed; the same local history appeared without that credential.

Keep the local history visible when this speculative Go request fails, while still reporting the API failure.

robinebers (changes requested):
I’m requesting changes because this can show incorrect spend totals and keep using an old account’s credentials. Migrated requests can be counted twice, while the model requests used to summarize long conversations are left out. Older OpenCode 2 history can also put paid OpenAI API-key requests into Codex subscription totals. Zen-only users can lose their local spend tiles when the Go service is unavailable.

The credential lookup needs to follow the account OpenCode currently uses: it can select both an inactive database credential and a stale key left in auth.json. Following the app’s advice to log in again will not reliably fix either case.

I found six actionable issues, detailed inline. Verification: the build succeeded; the original focused suite had 179 tests pass, one skipped, and no failures. Temporary checks against real SQLite credentials and message fixtures reproduced all six issues. Those checks were removed after the review; no source changes were made.

Reviewed commit: `2335ce79566a1aa0e8125c3a8c76b0fa134dc766`.

— **Rob’s AI Reviewer**

Sources/OpenUsage/Providers/OpenCode/OpenCodeUsageScanner.swift
arpankanwer: Fixed in f3079cc4 — and the local database says this was the biggest correctness issue in the PR: 20,607 hosted-filtered IDs exist in both tables, so the union was overcounting by roughly 40%.

Message IDs now ride through the projection as a sixth column and copies are deduplicated across tables and channel databases before accumulating, preferring the later timestamp then the larger token count (same rule the Codex scanner already used). Verified with same-ID-twice fixtures plus a same-row-on-two-databases case. Full suite: 1356 passed, 3 skipped, 0 failures.

Sources/OpenUsage/Providers/OpenCode/OpenCodeAuthStore.swift
arpankanwer: Fixed in f3079cc4. All four credential queries now use your ordering verbatim — `ORDER BY active DESC, time_updated DESC, id DESC` — with `(active IS NULL OR active = 1)` so imported NULL-flagged rows stay eligible without an unordered inactive row winning. The Codex check takes the current row first and only then tests `type = oauth`, so a live API key is no longer masked by a stale OAuth row.

Verified against a synthetic table (active = 0/1/NULL mix): the active row wins, NULL imports remain selectable, and the current-API-key-over-stale-OAuth case returns the key. A SQL-shape test pins the clause on all four queries.

Sources/OpenUsage/Providers/OpenCode/OpenCodeAuthStore.swift
arpankanwer: Fixed in f3079cc4 by resolving the source: the live database now wins, and `auth.json` is the fallback for OpenCode 1 (no credential table reads as absent, so v1 behavior is unchanged). Covered both directions in tests — stale file key plus live DB key returns the DB key, and a stale file OAuth claim no longer masks a live API-key row.

One residual edge worth naming: a v2 logout that removes the DB row while a stale file key remains will still surface the file key. Nothing local distinguishes that from a v1 install, so it keeps today's behavior rather than guessing.

Sources/OpenUsage/Providers/OpenCode/OpenCodeCodexUsageScanner.swift
arpankanwer: Fixed in f3079cc4. Both scanners accept `type = 'compaction'` with `status = 'completed'` as its own completion condition — confirmed on this machine that the assistant markers genuinely don't exist on those rows. The v1 branch stays assistant-only since that table has no such rows.

Note on magnitude here: 42 of my 43 local compaction rows carry no cost at all, so this changes almost nothing on older history and matters as newer builds record spend there. Covered by SQL-shape tests on both scanners plus a 151k-token counting test.

Sources/OpenUsage/Providers/OpenCode/OpenCodeCodexUsageScanner.swift
arpankanwer: Fixed in f3079cc4 with the credential-age bound: v2 `openai` rows older than the current OAuth credential's creation are excluded from Codex attribution; rows from its lifetime count. v1 rows priced paid traffic correctly and are unaffected, and file-based OAuth (no timestamp) keeps current behavior.

Two things I checked before choosing this over a schema marker: `seq` can't separate the eras (added June 2026, before the experimental window), and the experimental table is otherwise schema-identical, so credential age is the only per-row signal available. Known trade-off: reconnecting OAuth moves the bound forward and drops legit rows from the earlier OAuth window — undercounting those beats attributing paid history, but say the word if you'd rather bound it differently (e.g. earliest OAuth row rather than current).

Covered by a three-row timing test: pre-credential v2 excluded, post-credential v2 and pre-credential v1 both counted.

Sources/OpenUsage/Providers/OpenCode/OpenCodeAuthStore.swift
arpankanwer: Fixed in f3079cc4. A failed Go request now logs and falls through to the local scan instead of returning early; the API error surfaces only when there is nothing else to show, so the empty-database loud-failure tests still pass unchanged. You're right that our `opencode`-key fallback widened this — previously only Go keys reached the request, now any Zen key does too, which is exactly why the degradation path matters.

Covered by two new provider tests: connection failure and 401 with local data both keep the Today tile with no error category.

PR #1226, opened 2026-09-06
fix: discover Claude Swap accounts and isolate their usage

maddada:
## TL;DR

Fixes #1227.

Discover accounts saved by Claude Swap and merge matching Claude Code, Desktop, and Swap logins into one card per account and organization. Cards retain matching credential fallbacks and show the organization before the email.

## What was happening

- Saved Swap accounts were missing because discovery only inspected the default Claude Code login and Desktop organizations.
- The initial Swap integration could discard Desktop fallback credentials or omit Swap credentials from a Desktop-backed card.
- Email-only names made organizations sharing an email indistinguishable.

## What this changes

- Discover Swap metadata and retain stable account-and-organization card IDs.
- Keep matching default, session, Desktop, and read-only vault sources on merged cards; reject credentials for a different identity.
- Put the organization first in card names so long emails do not hide it.
- Preserve generation checks for in-flight login changes and source-scoped credential writes; never refresh or write vault credentials.
- Deduplicate copied/shared session history and filter spending by recorded ownership.
- Add overlap, fallback, layout/pin persistence, credential-change, source-write, and same-email/different-organization regression coverage.
- Preserve a separate default card when the login has an account UUID but no organization ID; do not infer a saved organization or count its spending on that card.
- Try live-capable matching credentials before limited-scope credentials, so a default login without `user:profile` cannot hide a working saved session.
- Merge current upstream `main` (`bb055e2`), preserving its ownership and nested workflow fixes alongside Swap history folders.

## Heads-up

- Discovery runs at launch. Restart after adding or removing saved accounts.
- Unattributed history remains excluded when multiple accounts are known. Broader SDK/Conductor attribution is outside this PR.

## Tests

Latest verification, 2026-09-13:

- Four new regression tests reproduced the review cases before the fixes and all passed afterward. They cover UUID-only default discovery, matching full-scope session usage, wrong-organization rejection, and history isolation.
- Full host `swift test`: 1,345 total, 1,342 passed, 3 skipped, zero failures.
- Focused `swift test --filter 'Claude|ProviderAccountAssemblyTests'`: 158 total, 156 passed, 2 skipped, zero failures.
- Rebuilt and signed the app with `CONFIG=debug ./script/build_and_run.sh build`, then launched the fresh host app. All four discovered Claude cards completed live refreshes successfully.
- `git diff --check`: passed.
[...]

PR #1264, opened 2026-09-15
Add Codex Swap account support

### LINKED ISSUE #1263: Discover Codex Swap accounts as separate usage cards
## Description

OpenUsage currently reads the default Codex login, but does not discover the other accounts saved by [Codex Swap (`xswap`)](https://github.com/maddada/codex-swap). Users with personal and work subscriptions need separate usage cards without changing the default login.

This follows the Claude Swap account discovery added in #1226.

## Expected Behavior

- Discover each saved Codex account and workspace, with distinguishable labels.
- Merge matching default, saved-home, and Keychain logins onto the same card.
- Keep card IDs, layout, pins, limits, and reset-credit actions attached to the correct identity when the default changes.
- Fall back only to a valid matching login. OpenUsage must never rotate or write Swap credentials.
- Exclude spending when the history cannot reliably identify its owner.

## Proposed Contribution

I am preparing a focused Codex Swap discovery PR, with regression coverage for overlapping sources, incomplete default identities, rejected/expired credentials, switching, stale responses, and shared history. Personal and work accounts are available in a Tart VM for live verification with concurrent `xswap run` sessions.

The PR will include before/after screenshots and the verification results. User documentation will stay brief; one-off QA logs will be kept in the PR, following the feedback on #1226.

maddada:
## TL;DR

Discover accounts saved by [Codex Swap (`xswap`)](https://github.com/maddada/codex-swap) as separate OpenUsage cards. Each account keeps its own limits, matching logins, layout, pins, and reset-credit action when the default Codex login changes.

Fixes #1263. Follows the account-discovery approach from #1226 and includes coverage for the issues raised during that review.

## What was happening

- OpenUsage followed the default Codex login and did not discover other accounts saved by Swap.
- Switching the default changed which account supplied the single card's limits.
- Codex's shared session history does not reliably identify which account paid for each turn, so it cannot safely be copied onto every account card.

## What this changes

- Read Swap's account registry, including custom `XSWAP_HOME` and `XDG_DATA_HOME` locations, without invoking aliases or modifying Swap's files.
- Discover Keychain-only defaults and custom Swap main homes, including accounts with no saved slot. Merge matching default, saved-file, and Keychain logins by workspace and user. Keep other identities separate and label cards with the Swap alias and email, or a workspace label when no alias is available.
- Preserve card IDs, existing layout and pins, and the existing Codex metric defaults. Bind reset-credit actions to the selected account.
- Try another matching access token after expiry or a 401/403. Discard responses when the supplying credential changes. Keychain discovery and credential rechecks run in the background. Swap account credentials are read-only: refresh tokens are neither used nor written.
- Keep partially identified default logins visible without borrowing a known account's credentials or sharing their history.
- Exclude unattributed local spending when multiple accounts are known, clearing cached spending before first display, CLI cache hits, and iCloud export even if refresh fails. Deduplicate shared/copied history for a single known account and match synced history by complete identity.
- Keep account-aware iCloud exports valid when Claude has no resolved identity, omitting only that history so other providers can sync.
- Update the user-facing Codex and iCloud documentation. Verification details stay in this PR rather than adding a one-off testing document.

## Heads-up

- Restart OpenUsage after adding, removing, or renaming Swap accounts. Existing cards continue refreshing across default-login changes.
- Codex owns token renewal for Swap accounts. If no matching access token works, the card asks the user to sign in again.
- `xswap switch` refuses to switch while Codex processes are running. OpenUsage respects that guard; concurrent `xswap run` sessions are supported.
- Broader attribution of shared Codex, SDK, or third-party session history is outside this PR's scope. Live subscription limits remain available for every account.

## Tests

GitHub [CI passed](https://github.com/robinebers/openusage/actions/runs/34960502013) for `5b7d471`. Host and Tart each passed the full suite: **1,374 passed, 3 skipped, 0 failed (1,377 total)**. The skips are the existing opt-in Claude/Codex local-history parity checks and live Claude test. Built using `CONFIG=debug ./script/build_and_run.sh build`; the arm64 app and compiled tests were transferred to and executed on macOS 15.7.7 in Tart.

Automated regression coverage in `CodexSwapAccountTests`, `CodexSwapHistoryTests`, `CodexSwapReviewRegressionTests`, `CodexSwapMaintainerReviewTests`, and `UsageReaderTests`:

| Prior review concern | Coverage |
| --- | --- |
| Overlapping sources and stable cards | Two account/workspace identities, repeated discovery, default switches, legacy-card migration, layout and pins, fresh-install defaults |
| Matching fallback and isolation | Expired tokens, both 401 and 403, saved-file and Keychain fallback, rejection of another account's credentials, no refresh-token use or credential writes |
| Credential changes during refresh | Default and saved Swap files replaced in flight; stale responses discarded and replacement identities cannot cross cards |
| Partial identity disappearing | Default login missing either workspace or email remains visible and cannot borrow the complete identity's credentials; the new regression failed before the fix |
| Distinguishable identities | Same email in different workspaces and different users in the same workspace remain separate |
| Spending ownership | Shared/copied rollouts deduplicated for one known account, unowned history excluded for multiple accounts, identity-based peer mapping, partial identities excluded from import/export |
| Findings in this PR review | All five reproduced as failing tests before the fix: mixed Claude/Codex/Grok export validation, Keychain-only discovery and stable IDs, background Keychain rechecks, removal of cached spending before refresh failure, custom main-home credential retention. Also covers Keychain replacement in flight and the CLI cache-hit path with a single-account control |
| Account-specific actions | Reset-credit lookup and claim stay with the selected account after a default switch, using fake HTTP responses |

Live Tart checks used the guest's **cua-driver CLI** with existing personal and work accounts:

- Both concurrent `xswap run --share-history` sessions reached normal Codex prompts. Dashboard and CLI refreshes kept the work Team and personal Pro 20x limits on separate cards, matching `xswap usage --all --json`.
- Hashes of the registry, main auth file, and both saved auth files stayed unchanged across OpenUsage refreshes.
- Isolated temporary homes with deliberately invalid preferred access tokens fell back to the same accounts' valid logins against the live API. Real credential files stayed unchanged.
- Confirmed Swap refused a global switch while the sessions ran. After exiting them normally, switched work to personal and back while OpenUsage remained running. Card IDs, plans, limits, layout, and pins stayed attached to their identities.
- Restored work as the default. `codex login status` still succeeded for all three homes, and both accounts still returned live usage. Relaunched the final app build and checked both cards again. The first live request timed out; a retry succeeded with limits matching xswap and no credential-file changes.

During the final check, a separate VM profile cleanup moved the default home into a backup. Its auth file still matched the original hash, both saved Swap logins were unchanged, and both cards continued refreshing from them.

After addressing this review, repeated the full suite in Tart and launched the rebuilt app through cua-driver. Both real Codex accounts returned the correct plans and limits; registry and saved-auth hashes stayed unchanged. A live isolated custom-main-home fixture also loaded the personal account without a saved slot alongside the saved work account. Its temporary credential copy contained no refresh token and was removed after the check.

Real tokens were not revoked or deliberately expired; those cases use controlled tests. No live reset credits were consumed. iCloud transport was not exercised; document export, validation, and aggregation were tested automatically.

## Screenshots

Captured directly from OpenUsage in Tart. Before uses the existing build without Codex Swap discovery; after uses this branch. The before capture uses work as the default; the after capture shows both saved accounts.

| Before: One Codex Card | After: Separate Work and Personal Cards |
| --- | --- |
| Before: only the default Codex Team account is visible | After: Codex work and personal accounts have separate cards and usage limits |


robinebers (changes requested):
**Rob’s AI Reviewer**

Reviewed commit `6788c2806b8aa01287478aeea0b24554118bde36`.

I found **five bugs worth fixing before merging**:

1. **iCloud history syncing can stop for every provider.** If Claude’s account identity is unavailable, Codex account cards cause the app to produce an invalid sync file. The entire upload fails, including other providers’ history. [Code](https://github.com/robinebers/openusage/blob/6788c2806b8aa01287478aeea0b24554118bde36/Sources/OpenUsage/Stores/WidgetDataStore.swift#L453)

2. **Keychain-only Codex accounts disappear.** When another account is saved in Swap, a working default login stored only in Keychain gets no card. Its usage limits become unavailable. [Code](https://github.com/robinebers/openusage/blob/6788c2806b8aa01287478aeea0b24554118bde36/Sources/OpenUsage/Services/ProviderAccountAssembly%2BCodex.swift#L45)

3. **Refreshing can freeze the interface.** Keychain credentials are repeatedly checked on the app’s main thread. Each check can block for five seconds, making the dashboard unresponsive. These checks need to run in the background. [Code](https://github.com/robinebers/openusage/blob/6788c2806b8aa01287478aeea0b24554118bde36/Sources/OpenUsage/Providers/Codex/CodexAuthStore%2BAccounts.swift#L25)

4. **Excluded spending can remain displayed and synced.** After adding a second account, old spending is cleared only after a successful usage refresh. If the first account’s login has expired, its cached spending remains—even though ownership is now considered unreliable. [Code](https://github.com/robinebers/openusage/blob/6788c2806b8aa01287478aeea0b24554118bde36/Sources/OpenUsage/Providers/Codex/CodexProvider%2BAccounts.swift#L37)

5. **Custom Swap homes can show a false login error.** The app discovers a valid default login in Swap’s custom main home, but forgets that location if the account has no saved Swap slot. The resulting card cannot load its credentials. [Code](https://github.com/robinebers/openusage/blob/6788c2806b8aa01287478aeea0b24554118bde36/Sources/OpenUsage/Services/ProviderAccountAssembly%2BCodex.swift#L69)

All five were reproduced with temporary tests. The existing full suite passed: **1,367 passed, 3 skipped**.

maddada:
@robinebers addressed all five findings in 5b7d4719c898238062665bafd75f552d0260d612:

1. **iCloud export:** account-aware documents omit unresolved Claude history while preserving Codex and other providers. The regression validates the full Claude/Codex/Grok document and preserves legacy single-Claude export behavior.
2. **Keychain-only accounts:** discovery now includes the default Keychain identity before saved Swap slots. It gets its own stable card and merges with matching file/Swap sources.
3. **Main-thread blocking:** Keychain discovery and every credential recheck run off the main actor. Tests record the thread used for each read and cover Keychain replacement during a request, including replacement by the other identity.
4. **Cached spending:** excluded local history and derived spend rows are stripped before app launch display or a CLI cache hit. This is persisted without changing live-limit timestamps, identity stamps, or cache freshness. Expired-login failures cannot retain or export the old spending; a single-account control keeps allowed history.
5. **Custom main homes:** cards retain the credential homes actually observed for their identity, even when that account has no saved Swap slot.

All five original findings were reproduced as failing regression tests before applying the fixes. Added seven tests overall. The full suite passed on both the host and Tart: **1,374 passed, 3 skipped, 0 failed (1,377 total)**. The three skips are existing opt-in tests.

Rebuilt and launched the app in Tart using the guest's cua-driver CLI. Refreshed the dashboard and CLI, checked both real account cards, and confirmed the registry and both saved credentials were unchanged. A live isolated custom-main-home test also loaded personal usage without a saved slot alongside the saved work account. Its temporary credential copy contained no refresh token and was removed afterward. Automated tests cover Keychain-only discovery and Keychain thread/isolation behavior without changing the real Keychain.

iCloud document export, validation, and aggregation were tested automatically; live iCloud transport was not exercised. The PR description and Codex/iCloud docs are updated. Ready for another review.

PR #1266, opened 2026-09-15
feat(codex): support multiple Codex accounts across Codex homes and pi logins

### LINKED ISSUE #1265: Support multiple Codex accounts across Codex homes and pi logins
I run two ChatGPT accounts through Codex. Work lives at `~/.codex` (and a `~/.codex-work` copy), personal at `~/.codex-personal`, and both are logged into pi as `openai-codex` and `openai-codex-2` in a pi-codex-multi pool. OpenUsage shows one Codex card: the work account's limits with the personal account's pi spend folded in. Yesterday that card read $703 of spend next to a weekly meter at 3% used. Two accounts, one card.

What I want is one card per ChatGPT account, matched by account id across Codex homes (`CODEX_HOME`, `~/.codex`, `~/.config/codex`, sibling `~/.codex-*` dirs) and pi's `auth.json` entries. The account at the default home keeps the plain `codex` id so nothing moves for existing installs. Codex home credentials keep refreshing the way they do today; pi credentials stay read-only since pi rotates its own tokens under a lockfile. Spend splits per card: each card reads its own homes' rollouts, and pi sessions route by pi provider id.

This is Phase 5a of `docs/research/account-first-plan.md` and the same shape as the Claude Swap discovery in #1226, except it covers the Codex CLI's own `CODEX_HOME` convention instead of a specific switcher tool (#1263 covers xswap). I have the branch working against my two accounts and will open the PR against this issue.

jal-co:
## Approved issue

Fixes #1265

## TL;DR

Every ChatGPT account signed into Codex on the Mac gets its own Codex card. Logins come from the Codex homes (`CODEX_HOME`, `~/.codex`, `~/.config/codex`, sibling `~/.codex-*` dirs) and from pi's `auth.json`, merged by account id.

## What was happening

- OpenUsage read one Codex login (the default home) and put every Codex spend source on that card.
- With two accounts, the work card's limits sat next to the personal account's pi spend. My card showed $703 spent against a weekly meter at 3% used.
- The login-shell capture never parsed on a machine whose `.zshrc` prints a banner (fastfetch), so the account pass never ran there at all.

## What this changes

- `CodexAccountDiscovery` finds every Codex home and pi `openai-codex*` login that can name its account. A login with no account id never becomes a card.
- `ProviderAccountAssembly` merges them into `CodexAccountCard`s. The default-home account keeps the bare `codex` id; others mint `codex@<hash8>` and seed the same layout defaults. Names come from the pi `multi-pass.json` label, else the email.
- Each `CodexProvider` is scoped to its card: its homes' `auth.json` files (refresh and rotate back as before), a read-only pi credential fallback, rollouts from its homes only, pi sessions filtered by pi provider id. Keychain and OpenCode sources stay on the default card. Reset-credit claims are per card.
- Sync documents carry Codex identities under the v2 schema so peer Codex history merges by account. Single-account Macs keep writing v1 files older builds accept.
- `LoginShellEnvironment.parse` matches the begin marker as a token suffix so a banner before it no longer breaks the capture.

```diff
 ~/.codex/auth.json        (acct A) ─┐
 ~/.codex-work/auth.json   (acct A) ─┼─▶ codex            "Codex: work"
 pi openai-codex-2         (acct A) ─┘
 ~/.codex-personal/auth.json (acct B) ─┐
 pi openai-codex             (acct B) ─┴─▶ codex@1b10751b  "Codex: you@example.com"
```

## Heads-up

- pi tokens are never refreshed by OpenUsage. A card that only has a pi login and an expired token reads "Token expired. Use this account in pi to refresh it."
- A keychain-mode Codex login still leaves the family as one card (plan phase 5b).
- Long account names truncate in the card header and the Total Spend legend, same as Claude account cards today. Worth a follow-up.
- Older builds reject v2 sync files that contain `codex@…` entries, the same step Claude multi-account took. Macs with one Codex account are unaffected.

## Tests

`swift build` and `swift test` with the Xcode toolchain: 1366 tests, 3 skipped. The two `ClaudeProviderTests` no-credentials tests fail on my machine on `main` as well (they read my real `~/.pi` logs). New coverage in `CodexMultiAccountTests` (discovery, assembly, id stability across a default-home swap, catalog scoping, pi expiry and 401 without refresh, pi spend split) plus a `LoginShellEnvironmentTests` regression for the banner case.

Live: built and ran the app against my two real accounts. Both cards refresh, the local API returns both under `/v1/usage/codex`, and the spend tiles split per account.

## Screenshots

Before is `main`; after is this branch. The popover sizes to its content, so the after frames are taller.

| | Light | Dark |
| --- | --- | --- |
| Before | One Codex card showing the work account's limits with the personal account's spend, light mode | One Codex card showing the work account's limits with the personal account's spend, dark mode |
| After | Two Codex cards, work and personal, each with its own limits and spend, light mode | Two Codex cards, work and personal, each with its own limits and spend, dark mode |


robinebers (changes requested):
**AI Review (Codex) — Request Changes**

Posted at the maintainer’s request after reviewing the approved issue, PR implementation, existing discussion, and current main.

Reviewed PR head `835265ec08b38fa24587db40ebbbecd6a7e46519` against remote `main` at `56378e5765f85d38ff413036fd984afe3d4664e4`, verified unchanged at the end of the review. The common ancestor is `86df736e`. This is a review of the submitted implementation and its compatibility with current main; no conflict-resolved merged implementation exists yet.

[Issue #1265](https://github.com/robinebers/openusage/issues/1265) is open, approved, and assigned to the PR author. Its additional discovery scope remains useful. The earlier policy-bot comment is stale; the current approved-issue check succeeds.

[PR #1264](https://github.com/robinebers/openusage/pull/1264) was merged after this branch diverged. Main already supports Codex Swap account cards, workspace/user identities, matching file and Keychain fallback, per-card reset claims, account layout defaults, and account-aware sync. It also protects against changed credentials, token writes to Swap snapshots, and ambiguous local/cached spend. Main does not yet provide #1265’s general sibling-home and pi-auth discovery.

The PR is marked DIRTY. A three-way comparison found conflict markers in 14 files. This requires an integration rewrite around main’s existing account model, beyond resolving text conflicts.

**Findings**

1. **[P1] Bind each credential read and action to the card’s expected identity.** [CodexProvider.swift:20](https://github.com/jal-co/openusage/blob/835265ec08b38fa24587db40ebbbecd6a7e46519/Sources/OpenUsage/Providers/Codex/CodexProvider.swift#L20)
   The card’s identity is discarded when constructing its auth store; only its launch-time file paths remain. A subsequent login switch makes those paths return another account’s credential. The original card then displays that account’s limits, and its reset service can consume that other account’s credit. Mocked tests reproduce both: A’s card sends B’s account header and successfully consumes B’s fake credit. A separate pending-refresh test also replaces auth.json with B’s login during token renewal; the PR overwrites that replacement with A’s refreshed credentials. Main’s identity filtering, current-credential rechecks, and read-only Swap handling prevent these cases. Preserve those protections, including checks around network awaits and source-specific write policy.

2. **[P1] Preserve main’s workspace-plus-user identity contract.** [ProviderAccountAssembly+Codex.swift:40](https://github.com/jal-co/openusage/blob/835265ec08b38fa24587db40ebbbecd6a7e46519/Sources/OpenUsage/Services/ProviderAccountAssembly+Codex.swift#L40)
   Discovery and reconciliation use the bare ChatGPT workspace/account id. Main’s Codex identity key is `accountID|email`, deliberately keeping two users in one workspace separate. The PR collapses that fixture into one card. Seeding a registry with main’s existing `acct-a|a@example.test` record then running PR discovery mints a new `codex@…` card for the same login, breaking the binding to its existing layout/pins. A valid main v2 document using the composite identity also fails to match the PR’s bare identity. Reuse CodexAccountIdentity and extend its validated credential parsing for the new sources.

3. **[P1] A home’s current login does not own its historical spending.** [ProviderAccountAssembly+Codex.swift:89](https://github.com/jal-co/openusage/blob/835265ec08b38fa24587db40ebbbecd6a7e46519/Sources/OpenUsage/Services/ProviderAccountAssembly+Codex.swift#L89)
   The PR assigns every rollout under a home to the account currently logged into that home. The regression fixture writes 120 tokens while A occupies the default home, then swaps only the auth files. B’s card receives those unchanged historical 120 tokens. Shared or copied session roots can also contribute to multiple cards because event deduplication is per scan/card. OpenCode’s unattributed history is always assigned to the current default, or to the first discovered identity when no default resolves. Main intentionally excludes these histories when multiple accounts are known and strips invalidated cached spend before display/export. Keep that exclusion unless durable ownership is proven. The same history caveat applies to pi provider IDs if their login identity is replaced or pool slots are reused.

4. **[P2] Keep the bare card’s identity mapping after the default changes.** [ProviderAccountAssembly+Codex.swift:96](https://github.com/jal-co/openusage/blob/835265ec08b38fa24587db40ebbbecd6a7e46519/Sources/OpenUsage/Services/ProviderAccountAssembly+Codex.swift#L96)
   In the swap fixture A permanently owns `codex`, while B permanently owns its hashed id. The loop correctly installs both identity mappings, then this cleanup deletes A’s `codex` mapping because B is now the default. That bypasses A’s cache ownership check and omits its history from multi-card export. The same assembly returns cards in current-default observation order, reversing their registry order after a switch. Main returns persistent record order. The PR’s existing swap test checks IDs and the default-source flag, but neither the ownership map nor order.

5. **[P2] Reload pi credentials when pi renews them.** [CodexAuthStore.swift:141](https://github.com/jal-co/openusage/blob/835265ec08b38fa24587db40ebbbecd6a7e46519/Sources/OpenUsage/Providers/Codex/CodexAuthStore.swift#L141)
   `loadPiAuth()` reconstructs auth from the PiCodexLogin snapshot captured at launch; it never rereads pi’s auth.json. Expiry checks also use that captured expiry. Renewing the token in pi and pressing Refresh therefore keeps returning authExpired with no HTTP request. This contradicts the new troubleshooting instruction to use the account in pi and refresh. Keep pi read-only, but reread the matching entry and validate its identity on each refresh.

6. **[P2] Retain all matching pi credentials for fallback.** [ProviderAccountAssembly+Codex.swift:91](https://github.com/jal-co/openusage/blob/835265ec08b38fa24587db40ebbbecd6a7e46519/Sources/OpenUsage/Services/ProviderAccountAssembly+Codex.swift#L91)
   All matching provider IDs are retained for spending, but only `pis.first` supplies credentials. With one expired `openai-codex` login and a valid `openai-codex-2` login for the same account, the card reports authExpired without trying the valid login. Preserve the candidate set and try matching usable access tokens without rotating pi credentials.

7. **[P2] Do not strip ownership based on which cards have history today.** [WidgetDataStore.swift:442](https://github.com/jal-co/openusage/blob/835265ec08b38fa24587db40ebbbecd6a7e46519/Sources/OpenUsage/Stores/WidgetDataStore.swift#L442)
   Export decides account awareness from the histories actually inserted, remaps a single exported account to the bare family id, and drops all Codex identities if no hashed provider remains. Two registered Codex accounts with only A’s history export an unowned v1 document. A peer with two matching cards then imports none of A’s spending. A single-account peer can instead accept unowned spending without proving an account match. Main retains complete Codex ownership even when only one account exports. Preserve its export/import identity rules.

8. **[P2] Keep account-aware exports valid when Claude is unresolved.** [WidgetDataStore.swift:447](https://github.com/jal-co/openusage/blob/835265ec08b38fa24587db40ebbbecd6a7e46519/Sources/OpenUsage/Stores/WidgetDataStore.swift#L447)
   Two Codex histories cause v2 export, but an unresolved bare Claude history is still included without its required identity. The produced document throws `invalidIdentity("claude")`, blocking the whole device’s history export. Main explicitly omits only unresolved Claude history so other providers can sync. Preserve that behavior.

9. **[P2] Make default identity observation and discovery use compatible parsing.** [ProviderAccountAssembly+Codex.swift:22](https://github.com/jal-co/openusage/blob/835265ec08b38fa24587db40ebbbecd6a7e46519/Sources/OpenUsage/Services/ProviderAccountAssembly+Codex.swift#L22)
   Discovery accepts an access-token JWT identity, while the unchanged DefaultAccountObserver only checks the id_token and stored account_id. An access-token-only default login yields unresolved, and this early return suppresses even independently identifiable sibling/pi cards. A controlled fixture produces zero cards instead of two. Comma-separated CODEX_HOME is also split by discovery but treated as one literal path by the observer, so the promised list support cannot reliably establish a default-source owner. The discovery code’s payload fallback is only used when the id_token cannot be decoded; a decodable id_token missing the claim does not trigger access-token fallback.

10. **[P2] Carry a discovered JWT account id into usage/reset request credentials.** [CodexAccountDiscovery.swift:95](https://github.com/jal-co/openusage/blob/835265ec08b38fa24587db40ebbbecd6a7e46519/Sources/OpenUsage/Providers/Codex/CodexAccountDiscovery.swift#L95)
    A login with an id_token account claim but no stored tokens.account_id creates a correctly identified card. The auth store then returns the original unmodified auth, and usage/reset requests omit ChatGPT-Account-Id. A mocked request test reproduces the missing header. Main’s scoped auth store fills the selected workspace into the request auth state. Discovery and request routing must use the same resolved identity.

**Existing main issue, rather than a new PR regression**

The Copilot comment about CodexPricingSection is valid behaviorally: every Codex card exposes the global fallback-model preference, but the wait/backoff/refresh code still targets literal `codex`. Those view files are identical on current main and PR head. Track this as an existing main follow-up. Because the preference is global, recalculation should reach every applicable Codex card; changing only the selected card’s refresh id would still leave other cards stale.

The empty pi-ID scan can also produce a misleading pi source note. It is a smaller presentation issue, not evidence of token/spend leakage by itself. Skip empty scoped scans or derive the note from matching entries.

**Verification**

| Scope | Result |
| --- | --- |
| PR’s new multi-account, shell, aggregation, and document suites | 36 passed |
| Existing affected PR auth, resets, native/pi scanners, account registry/cache, fallback pricing, and CLI suites | 104 passed, 1 skipped |
| Current main’s Codex Swap, history, shell, aggregation, and document suites | 43 passed |
| Additional isolated review reproducers | 15 failed tests, 21 assertion failures reproducing the findings above |

Both isolated revisions compiled through `swift test`. A local Xcode beta/Sparkle test-runner setup issue was resolved before executing the suites. Review reproducers were added only to the temporary review checkout; repository source was unchanged.

These are focused suites, not a new full-suite run. GitHub’s CI run reports failure but contains zero jobs and zero check runs; it provides no build/test result to interpret. A fresh Build and Test run is needed after rebasing. The current approved-issue check is successful.

All account requests and reset consumption in review reproducers use fake HTTP clients. Real credentials were not edited and no live credit was consumed. iCloud transport and a fresh live UI session were not exercised. The supplied screenshot was inspected; it documents the intended cards but cannot verify credential/history ownership.

**Recommended revision**

Rebase onto current main and add sibling-home/pi discovery as additional source types. Keep xswap discovery, existing composite identities and record IDs, background Keychain reads, matching-only fallback, read-only Swap/pi credentials, current-credential checks, cache history exclusion, and main’s valid account-aware sync export. Add live pi token reload and complete fallback candidates. Do not claim per-account historical spending merely from the current home or provider-slot assignment. Keep the shell-banner parser fix: it is small, independently tested, and still absent from main. Re-run the existing Codex Swap regression suites alongside the new source tests and obtain fresh CI after integration.

jal-co:
@robinebers fixed, re requested

robinebers:
@jal-co thank you

@pullfrog review

robinebers:
@pullfrog review

jal-co:
@pullfrog review

jal-co:
@robinebers should all be fixed, tyvm

PR #1271, opened 2026-09-17
test: isolate Codex local-spend test from real pi/OpenCode history

### LINKED ISSUE #1269: CodexProviderTests local-spend test is not isolated from real pi/OpenCode history
**TL;DR** — `CodexProviderTests.testNoUsageDataBadgeIsDroppedWhenLocalLogsHaveSpend` (`Tests/OpenUsageTests/CodexProviderTests.swift:493`) injects only the Codex rollout scanner, but `CodexProvider.snapshot` also merges `PiUsageScanner.shared` and the default `OpenCodeCodexUsageScanner`. On a machine with local OpenCode (or pi) Codex history, the snapshot folds that real spend in, so the test fails on every run regardless of the code under test.

### What happens

`CodexProvider.snapshot(mapped:)` merges three local sources:

- the injected `logUsageScanner`
- `PiUsageScanner.shared` (`Sources/OpenUsage/Providers/Codex/CodexProvider.swift:166`)
- `openCodeUsageScanner`, which defaults to the real `OpenCodeCodexUsageScanner()` (`CodexProvider.swift:171-172`)

The test injects only the first (`CodexProviderTests.swift:513`):

```swift
logUsageScanner: CodexLogFixture.scanner(home: home),
```

so `openCodeUsageScanner` stays on the real default and reads the developer's actual OpenCode database. The assertion then compares the fixture's `0.25` / `150 tokens` against the machine's real totals. Observed locally:

```text
number: 1212.188..., kind: dollars   (expected 0.25)
number: 6755459.0,  kind: tokens     (expected 150)
```

### Repro

On a machine with Codex usage in the local OpenCode database (`~/.local/share/opencode/opencode*.db`):

```sh
swift test --filter testNoUsageDataBadgeIsDroppedWhenLocalLogsHaveSpend
```

Fails. Pointing the OpenCode data dir away makes it pass, which confirms the source:

```sh
OPENCODE_DATA_DIR=/nonexistent swift test --filter testNoUsageDataBadgeIsDroppedWhenLocalLogsHaveSpend
```

### Suggested fix

The sibling test already isolates itself by injecting a fake scanner (`CodexProviderTests.swift:569`):

```swift
openCodeUsageScanner: openCodeScanner,
```

For this test, the simplest complete isolation is `allowsUnattributedHistory: false`, which skips both unattributed sources (`CodexProvider.swift:166` and `:171`):

```swift
logUsageScanner: CodexLogFixture.scanner(home: home),
allowsUnattributedHistory: false,
```

I applied that locally and the test passes with the real database present. Happy to send it as a PR.

### Deeper cause

`PiUsageScanner.shared` is a non-injectable singleton, so tests can only dodge real pi history indirectly — the sibling test uses a far-future `now` for exactly that reason. Making it injectable would let these tests isolate deterministically instead of depending on the clock and on the developer's machine having no local history.

manelpb:
## TL;DR

`testNoUsageDataBadgeIsDroppedWhenLocalLogsHaveSpend` injected only the Codex rollout scanner, so on a developer machine with real OpenCode/pi Codex history the snapshot folded that real spend in and the test failed on every run. It now passes `allowsUnattributedHistory: false` so the test sees only its own fixture.

## What was happening

- The test injected `CodexLogFixture.scanner(home:)` but left the default `CodexProvider` initialization, which also merges `PiUsageScanner.shared` and the default `OpenCodeCodexUsageScanner`.
- On a machine with local OpenCode/pi Codex usage, `snapshot` folded that real spend into the result, so the `Today` assertion compared multi-million-token real spend against the fixture's `$0.25 / 150 tokens` and failed.
- Reproduced on a machine with local history: pre-fix the assertion got `$1212.19 / 6,755,459 tokens` where it expected `$0.25 / 150 tokens`. With both local sources empty the test passed, and each source separately changed the expected numbers — matching the reproduction in #1269.

## What this changes

- Passes `allowsUnattributedHistory: false` to the provider in this test, so the snapshot includes only the injected rollout fixture.
- No production code changes. The native Codex fixture and the "No usage data" badge regression check are unchanged.

## Heads-up

Per the issue, making the pi scanner injectable and removing the sibling test's far-future-date workaround are intentionally left as a separate follow-up.

## Tests

- `swift test --filter CodexProviderTests/testNoUsageDataBadgeIsDroppedWhenLocalLogsHaveSpend`
- Fails before this change on a machine with local OpenCode/pi Codex history; passes after.

## Screenshots

Not a visual change — test-only.

Fixes #1269


robinebers (approved):
Thanks, @manelpb, for tracking this down and isolating the test from local pi/OpenCode history. The change preserves the rollout fixture and badge regression assertions, and the focused test passed. Looks good!

